PR #1, opened 2021-07-25
feat: API scaffold with tests and terraform

maxneuvians:
This PR adds the basic scaffold for a API container that can be invoked through an API gateway event. It also includes the terraform required to stand up the infrastructure. The API is structured in such a way that it can handle API Gateway events using Fast API but also handle incoming events from SNS topics and S3 events.


.github/workflows/ci_python.yml
```diff
@@ -0,0 +1,53 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2 
+
+      - uses: dorny/paths-filter@v2.2.1
+        id: filter
+        with:
+          filters: |
+            package:
+              - '${{ matrix.package }}/**'
```
CalvinRodo, line 23: Instead of having each of the subsequent steps use an if statement to determine if they need to run we can have this step occur in another job and use the needs action to determine if we want to run this, although that might break down if we do have more packages in the matrix

https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idneeds

.github/workflows/tf_apply.yml
```diff
@@ -0,0 +1,50 @@
+name: "Terraform"
+
+on:
+  push:
+    branches:
+      - main
+
+env:
+  TERRAFORM_VERSION: 0.14.2
```
CalvinRodo, line 9: The devcontainer has TF v 1.0.3 installed and this one has 0.14.2 we should probably run new infra on v1+

.github/workflows/tf_apply.yml
```diff
@@ -0,0 +1,50 @@
...
+  TERRAGRUNT_VERSION: v0.29.2
```
CalvinRodo, line 10: TG is on v0.31.1

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,54 @@
...
+			// Update 'VARIANT' to pick a Python version: 3, 3.6, 3.7, 3.8, 3.9
+			"VARIANT": "3.9",
+			// Options
+			"INSTALL_NODE": "true",
+			"NODE_VERSION": "lts/*",
+			"SHELLCHECK_VERSION": "0.7.2",
+			"SHELLCHECK_CHECKSUM": "70423609f27b504d6c0c47e340f33652aea975e45f312324f2dbf91c95a3b188",
+			"TERRAFORM_VERSION": "1.0.3",
+			"TERRAFORM_CHECKSUM": "99c4866ffc4d3a749671b1f74d37f907eda1d67d7fc29ed5485aeff592980644",
+			"TERRAGRUNT_VERSION": "0.29.2",
```
CalvinRodo, line 18: TG is on version 0.31.1

.github/workflows/tf_plan.yml
```diff
@@ -0,0 +1,48 @@
...
+  pull_request:
+    paths:
+      - "terragrunt/**"
+      - ".github/workflows/**"  
+env:
+  AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+  AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+  AWS_REGION: ca-central-1
+  TERRAFORM_VERSION: 0.14.2
+  TERRAGRUNT_VERSION: v0.29.2
```
CalvinRodo, line 13: Same version issue as above.

.github/workflows/build_and_push.yml
```diff
@@ -0,0 +1,69 @@
...
+      - name: Logout of Amazon ECR
+        if: ${{ steps.filter.outputs.image == 'true' }}
+        run: docker logout ${{ steps.login-ecr.outputs.registry }}
+
+      - name: Deploy lambda
+        if: ${{ steps.filter.outputs.image == 'true' }}
+        run: |
+          aws lambda update-function-code \
+            --function-name ${{ matrix.image }} \
+            --image-uri $REGISTRY/${{ matrix.image }}:latest
```
CalvinRodo, line 69: Is this needed or will the lambda grab the image tagged latest when it runs next and just get the newest version?

terragrunt/aws/api/lambda.tf
```diff
@@ -0,0 +1,30 @@
...
+  //  variables = {}
+  // }
+
+  role = aws_iam_role.api.arn
+
+  lifecycle {
+    ignore_changes = [
+      image_uri,
+    ]
+  }
```
CalvinRodo, line 21: Do we expect this to change if we keep it at latest it should always be the same.

CalvinRodo (changes requested):

.github/workflows/build_and_push.yml
maxneuvians: This one is needed - it does not autoupdate

terragrunt/aws/api/lambda.tf
maxneuvians: `--image-uri $REGISTRY/${{ matrix.image }}:latest` actually updates it to the last SHA not the latest, which is odd - I added this in when the TF Plan threw up the difference

.github/workflows/tf_plan.yml
maxneuvians: Good catch - fixed

.devcontainer/devcontainer.json
maxneuvians: Good catch - fixed

.github/workflows/tf_apply.yml
maxneuvians: Good catch - fixed

.github/workflows/tf_apply.yml
maxneuvians: Good catch - fixed

.github/workflows/ci_python.yml
maxneuvians: Nice works - https://github.com/cds-snc/scan-websites/actions/runs/1067773758

mohamed-cds (approved):
Nicely organized. Looking forward to contributing! 🚀 ✅

CalvinRodo (approved):

PR #2, opened 2021-07-26
fix: correctly set dirs for CD on deploy and tf apply

maxneuvians:
Fixes a directory problem in the continuous deployment for tf_apply and pushing the lambda containers


mohamed-cds (approved):
🚀 ✅

PR #8, opened 2021-07-27
feat: add organisations model to show ORM and DB connector

### LINKED ISSUE #7: Add migrations manager and ORM
We will be using Postgres for our data and need something that can manage migrations and ideally an object-relational mapping. Standard for python seems to be SQLAlchemy and its ORM. Here is a tutorial for integration with FAST API, although there is nothing special about the two. https://fastapi.tiangolo.com/tutorial/sql-databases/

maxneuvians:
Closes #7 by scaffolding a basic ORM and DB connector using an `organisations` model as an example. Includes the required code inside docker-compose to do development against the database and to run tests locally.

To be added is a terraform module to deploy RDS + Proxy inside the AWS environment.


api/db_migrations/versions/06b310a6251e_create_organisations_table.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+def upgrade():
+    op.create_table(
+        "organisations",
+        sa.Column("id", postgresql.UUID(as_uuid=True), primary_key=True),
+        sa.Column("name", sa.Unicode(255), nullable=False),
+        sa.Column("created_at", sa.DateTime, default=sa.func.now()),
+        sa.Column("updated_at", sa.DateTime, onupdate=sa.func.utc_timestamp()),
+    )
```
dj2, line 27: Should there be an index on `name` as it's referenced in the py file as indexed?

api/models/organisations.py
```diff
@@ -0,0 +1,29 @@
...
+from sqlalchemy.ext.declarative import declarative_base
+
+Base = declarative_base()
+
+
+class Organisation(Base):
+    __tablename__ = "organisations"
+
+    id = Column(UUID(as_uuid=True), primary_key=True, default=uuid.uuid4)
+    name = Column(String, nullable=False, index=True, unique=False)
```
dj2, line 15: Should the name be unique? What does it mean to have two organizations with the same name?

api/db_migrations/versions/06b310a6251e_create_organisations_table.py
maxneuvians: I used the generator incorrectly :( - should be fixed now

api/models/organisations.py
maxneuvians: see above

.devcontainer/docker-compose.yml
```diff
@@ -0,0 +1,62 @@
...
+    image: postgres:11.2
+    volumes:
+    - ./initdb:/docker-entrypoint-initdb.d
+    restart: always
+    command:
+      - "postgres"
+      - "-c"
+      - "listen_addresses=*"
+    environment:
+      POSTGRES_USER: postgres
```
dj2, line 59: Is using the postgres user to get around creating a new user account in the db? Would it make sense to have a `scan-websites` user here?

.github/workflows/ci_python.yml
```diff
@@ -17,6 +17,16 @@ jobs:
 
   python-tests:
     runs-on: ubuntu-latest
+    services:
+      postgres:
+        image: postgres:11.8
+        env:
+          POSTGRES_USER: postgres
+          POSTGRES_PASSWORD: postgres
```
dj2, line 25: Is this a github setting? Or should it match the config from above?

api/Makefile
```diff
@@ -1,6 +1,6 @@
 .PHONY: dev
 dev:
-	uvicorn main:app --reload
+	uvicorn main:app --reload --host 0.0.0.0 --port 8000
```
dj2, line 3: Why `0.0.0.0` and not `127.0.0.1` to contain it to localhost?

api/models/organisations.py
```diff
@@ -0,0 +1,36 @@
+import datetime
```
dj2, line 1: Should the file be named `Organization` to match the class name?

api/models/organisations.py
```diff
@@ -0,0 +1,36 @@
...
+from sqlalchemy import DateTime, Column, String
+from sqlalchemy.dialects.postgresql import UUID
+from sqlalchemy.ext.declarative import declarative_base
+from sqlalchemy.orm import validates
+
+
+Base = declarative_base()
+
+
+class Organisation(Base):
```
dj2, line 13: `Organization` ? Weird word, English is weird, heh. The plural noun is apparently `Organisations` so the table name is right.

api/tests/models/test_organisations.py
```diff
@@ -0,0 +1,51 @@
...
+
+def test_organisation_model_saved(session):
+    organisation = organisations.Organisation(name="foo")
+    session.add(organisation)
+    session.commit()
+    assert organisation.name == "foo"
+    assert organisation.id is not None
+    assert organisation.created_at is not None
+    assert organisation.updated_at is None
+    session.delete(organisation)
```
dj2, line 19: Other tests also `rollback` do we need that here?

api/tests/models/test_organisations.py
```diff
@@ -0,0 +1,51 @@
...
+    session.commit()
+
+
+def test_organisation_empty_name_fails(session):
+    organisation = organisations.Organisation()
+    session.add(organisation)
+    with pytest.raises(Exception):
```
dj2, line 26: Can we validate that this is the right exception based on name being blank?

api/tests/models/test_organisations.py
```diff
@@ -0,0 +1,51 @@
...
+    assert organisation.name == "foo"
+    assert organisation.id is not None
+    assert organisation.created_at is not None
+    assert organisation.updated_at is None
+
+    organisation_two = organisations.Organisation(name="foo")
+
+    session.add(organisation_two)
+
+    with pytest.raises(Exception):
```
dj2, line 46: Ditto check this is a duplicate name exception?

.devcontainer/docker-compose.yml
maxneuvians: The docker container will just create a new db based on the `POSTGRES_USER`, it just makes it slightly more convenient to use the default user with tools like psql. It also removes some of the magic of that DB being auto-created by being explicitly defined in `POSTGRES_DB`

.github/workflows/ci_python.yml
maxneuvians: I matched the dev container to the CI container postgres as the password as well, again it can be anything

api/Makefile
maxneuvians: You can't access the dev server from outside the container in this case

api/models/organisations.py
maxneuvians: This is where conventions are tripping me - ideally you would pool all associated code for `organisations` in one file, including the main `Organisation` class so you can do weird stuff like:

```
from model.organisations import Organisation, OrganisationRelatedFunctionOrClass etc.
```

Let me know if you have any strong opinions

api/models/organisations.py
maxneuvians: Let us stay consistent, I like the S

api/tests/models/test_organisations.py
maxneuvians: `rollback` does not actually undo the transaction once commited, it will just rollback a broken transaction. At least that was my understanding.

api/tests/models/test_organisations.py
maxneuvians: Changed to `from sqlalchemy.exc import IntegrityError`

api/tests/models/test_organisations.py
maxneuvians: Changed to `from sqlalchemy.exc import IntegrityError`

api/models/organisations.py
dj2: I'd be inclined to name it to the singular.  Things like `Organisation.all()` make sense as the singlar.

api/Makefile
dj2: Hm, so this is only intended to run inside docker? I was thinking this could also be run outside docker

api/tests/models/test_organisations.py
dj2: Do you actually need to commit or can you just do the tests inside the transaction and then roll it all back? Is it doing validation on commit?

maxneuvians: The commit is what assigns the UUID and presumably executes the actually things that might fail in DB integrity. I

api/Makefile
maxneuvians: This would be following the dev container pattern we have been using through out CDS products. I would strongly recommend using docker containers to do dev as some of the dependencies (postgres, redis) are going to be much easier to manage inside docker compose than on your host machine. You also avoid the whole mess that is pythons virtual environments if you just run it in an isolated instance.

api/models/organisations.py
maxneuvians: Makes sense, changed

dj2 (approved):

PR #10, opened 2021-07-28
docs: Fast API ADR

### LINKED ISSUE #3: Write ADR for FAST API
We need an architecture decision record for why we are using FAST API

maxneuvians:
Closes #3 by adding an architecture decision record for why we are using Fast API


dj2 (approved):

CalvinRodo (approved):
:taco::cat:

PR #21, opened 2021-07-30
feat: adds axe-core scanner

### LINKED ISSUE #13: Create axe-core / puppeteer container
We need to create a container that can be invoked through a lambda which does the following:

- handles an invocation from a SNS topic
- visits a URL in puppeteer
- scans it for accessibility issues
- saves the accessibility report to an S3 bucket
- saves a screenshot of the page to an S3 bucket

Depending if we are using a python crawler or not, this function may also have to do a basic crawl for links. 

Our current implementation looks like this: https://github.com/cds-snc/a11y-tools/blob/main/processing/chromium-lambda/src/impl.ts

maxneuvians:
Closes #13. A duplicate of the code https://github.com/cds-snc/a11y-tools/blob/main/processing/chromium-lambda/src/impl.ts with the exception that it provides it's own chromium later binary and runs it in a docker container that can be invoked from a lambda. Also adds a CI function to lint and test the node code.

Terraform to deploy the lambda is added in another PR.


scanners/axe-core/src/index.ts
```diff
@@ -0,0 +1,146 @@
+import AxePuppeteer from "@axe-core/puppeteer";
+import { S3Event, SNSEvent } from "aws-lambda";
```
dj2, line 2: Out of curiousity, why remove the Impl as we had with the previous version? This appers to have the same issue where you have to jump through a bunch of mock hoops to make testable?

maxneuvians: Working my way back to it - just wanted to simplify it for now so I could make sure the docker configuration was working properly

scanners/axe-core/src/impl.ts
```diff
@@ -0,0 +1,154 @@
...
+import { BlobStore } from "./common/blobstore";
+import { MessageBus } from "./common/message_bus";
+
+const USER_AGENT =
+    "Mozilla/5.0 (CDS-SNC A11Y Tools; Puppeteer) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/89.0.4389.90 Safari/537.36";
+
+export async function Impl(
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    records: any[],
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any, @typescript-eslint/explicit-module-boundary-types
```
dj2, line 12: Some of these we should probably fixup as well. We know what methods we call on Browser and on the records so we should be able to create interfaces for them. We should work towards getting rid of all these `any`s

scanners/axe-core/src/impl.ts
```diff
@@ -0,0 +1,154 @@
...
+        return error;
+    } finally {
+        if (browser !== null) {
+            await browser.close();
+        }
+    }
+    return true;
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
dj2, line 53: This shouldn't be needed as there is no any on the next line

scanners/axe-core/src/impl.test.ts
```diff
@@ -0,0 +1,221 @@
...
+
+describe("Impl", () => {
+    beforeAll(() => {
+        Date.now = jest.fn(() => 1609477200);
+        process.env.REPORT_DATA_BUCKET = "foo";
+        process.env.SCREENSHOT_BUCKET = "bar";
+    });
+
+    afterEach(() => {
+        delete process.env.URL_SCAN_TOPIC;
```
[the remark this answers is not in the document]
maxneuvians: Removed

scanners/axe-core/src/impl.ts
maxneuvians: Removed

scanners/axe-core/src/impl.test.ts
```diff
@@ -0,0 +1,221 @@
...
+
+const msgBusMock = {
+    publish: jest.fn().mockReturnThis(),
+    promise: jest.fn(),
+};
+
+describe("Impl", () => {
+    beforeAll(() => {
+        Date.now = jest.fn(() => 1609477200);
+        process.env.REPORT_DATA_BUCKET = "foo";
```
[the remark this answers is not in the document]
maxneuvians: Should be in the next push

scanners/axe-core/Dockerfile
```diff
@@ -0,0 +1,84 @@
+# Define global args
```
dj2, line 1: So, this has a separate Dockerfile, does that mean I have to run it in a different container from the api? Is there a way to run them together locally? Or, do we run both containers locally and they talk together (is this a thing VSCode can do?)

maxneuvians: Yes, we could run them locally using express the way we ran them before in the old project. The other way is in VSCode you can run them in docker compose, the same way the API and two postgres containers are running in a local network.

maxneuvians: The hardest part will be how to bridge the two services, currently we are planning on using an SNS queue in the cloud, but locally that may be harder

dj2: That was fine in the old express app.  I created a fake `queue` which just wrote to disk, the other side would read all the files from the queue folder.

dj2 (approved):

PR #29, opened 2021-08-03
feat: add a healthcheck endpoint

CalvinRodo:
# Summary | Résumé

- Refactor database outside of main.py
- Refactor out logger
- Add a healthcheck that will return able to connect to the database


api/database/db.py
```diff
@@ -0,0 +1,7 @@
+import os
+
+from sqlalchemy import create_engine
+from sqlalchemy.orm import sessionmaker
+
+db_engine = create_engine(os.environ.get("SQLALCHEMY_DATABASE_URI"))
```
dj2, line 6: Does this keep the session alive? Create a new one per-connection? Are they pooled?

api/api_gateway/api.py
```diff
@@ -1,9 +1,30 @@
...
 
 
 @app.get("/version")
 async def version():
     return {"version": environ.get("GIT_SHA", "unknown")}
+
+
+def healthcheck_session(sessionmaker):
+    session = sessionmaker()
+    return session.query(text("1")).from_statement(text("SELECT 1")).all()
```
dj2, line 18: Out of curiosity (and not knowing sqlalchemy), can this just be something like `select.query("SELECT 1").all()`?

CalvinRodo: I've changed this I'm going to return the db migration version instead of random sql. 

And to answer your actual question I have no idea. I also don't know sqlalchemy.

api/database/db.py
CalvinRodo: I'm assuming this works in a similar way to how javascript lambdas work. 

The session will stay alive assuming we don't close it in the handler, which I may still do. 

It will potentially reshare the connection depending on if the container is re-run which sometimes happens.

dj2: @maxneuvians didn't we have to do a connection pool thing in the other lambdas in order to not run over some limit, or hit some issue with startup time?

dj2 (approved):

api/database/db.py
CalvinRodo: The RDS Proxy should handle connection pooling that the RDS module creates should handle this for us. 
https://github.com/cds-snc/terraform-modules/blob/eff1249f50f626f1bdb5207496c8832a46cd3645/rds/main.tf#L81

PR #32, opened 2021-08-04
feat: add a database connection timeout

CalvinRodo:
Postgresql connect_args documentation is here: https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-PARAMKEYWORDS

### DIFF
--- a/api/database/db.py
+++ b/api/database/db.py
@@ -3,5 +3,6 @@ import os
 from sqlalchemy import create_engine
 from sqlalchemy.orm import sessionmaker
 
-db_engine = create_engine(os.environ.get("SQLALCHEMY_DATABASE_URI"))
+connection_string = os.environ.get("SQLALCHEMY_DATABASE_URI")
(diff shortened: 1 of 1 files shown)


api/database/db.py
```diff
@@ -3,5 +3,6 @@
 from sqlalchemy import create_engine
 from sqlalchemy.orm import sessionmaker
 
-db_engine = create_engine(os.environ.get("SQLALCHEMY_DATABASE_URI"))
+connection_string = os.environ.get("SQLALCHEMY_DATABASE_URI")
+db_engine = create_engine(connection_string, connect_args={"connect_timeout": 10})
```
dj2, line 7: nit: Can you add a comment saying this is a `10 second timeout` to make it clear what the number means without searching for the docs?

dj2 (approved):

api/database/db.py
CalvinRodo: yes I can

PR #33, opened 2021-08-04
Minor shuffling and cleanup.

dj2:
# Summary | Résumé

This PR moves the Makefile to the top level so you don't need to search
for it. The postgresql-client is added to the docker container. A VSCode
tasks file is added to run the various make commands.


dj2:
This isn't going to work well with the docker files. Will go a different way in another PR.

PR #44, opened 2021-08-05
feat: add db migrations

### LINKED ISSUE #30: Add DB migrations to CD
Currently we don't have an automatic way of running migrations in prod - as a result our new `/healthcheck` endpoint does not work. One thought I had was to add a new step to the following CD workflow after this one:

https://github.com/cds-snc/scan-websites/blob/main/.github/workflows/build_and_push.yml#L67-L71

that basically does something like this

```
if ${{ matrix.image }} == 'api'
then
aws lambda invoke \
    --function-name api \
    --invocation-type Event \
    --payload '{ "task": "migrate" }' \
    response.json
```

which will tell a single instance of the lambda container to perform the migration.  That way you don't have a race condition to migrate because you know you are only triggering it on one container. 

Ideally you would be able to invoke the lambda with a container command that overriders the default `ENTRYPOINT` for the container, which you could then run `make migrations` with, but I don't think AWS is that far with containers in lambdas.

CalvinRodo:
# Summary | Résumé

- Watches for a custom migrate task and runs the db migration if found
- Adds a step to the build and push workflow to run a migration after
  the lambda is deployed
- Adds a script to test the repsonse of that migration invocation
- Adds a workflow that will run shellcheck and some tests on our scripts

Closes #30

### DIFF
--- a/.github/workflows/build_and_push.yml
+++ b/.github/workflows/build_and_push.yml
@@ -71,6 +71,7 @@ jobs:
             --image-uri $REGISTRY/${{ matrix.image }}:latest
 
       - name: Migrate Database
+        if: ${{ matrix.image == 'api' }}
         run: |
           aws lambda invoke \
             --function-name api \
--- a/.github/workflows/build_and_push.yml
+++ b/.github/workflows/build_and_push.yml
@@ -69,3 +69,12 @@ jobs:
           aws lambda update-function-code \
             --function-name ${{ matrix.image }} \
             --image-uri $REGISTRY/${{ matrix.image }}:latest
+
+      - name: Migrate Database
+        run: |
+          aws lambda invoke \
+            --function-name api \
+            --invocation-type Event \
+            --payload '{ "task": "migrate" }' \
+            response.json
+          .github/workflows/scripts/test_migration_resp.sh response.json
--- a/api/main.py
+++ b/api/main.py
@@ -1,6 +1,7 @@
 from mangum import Mangum
 from api_gateway import api
 from logger import log
+from database.migrate import migrate_head
 
 
 app = api.app
@@ -14,6 +15,14 @@ def handler(event, context):
         asgi_handler = Mangum(app)
         response = asgi_handler(event, context)
         return response
+
+    elif "migrate" in event:
+        try:
+            migrate_head()
+            return "Success"
+        except Exception as err:
+            log.error(err)
+            return "Error: check cloudwatch logs for error."
 
     else:
         log.warning("Handler recieved unrecognised event")
--- a/api/main.py
+++ b/api/main.py
@@ -16,7 +16,7 @@ def handler(event, context):
         response = asgi_handler(event, context)
         return response
 
-    elif "migrate" in event:
+    elif event["task"] == "migrate":
         try:
             migrate_head()
             return "Success"
--- a/api/main.py
+++ b/api/main.py
@@ -16,13 +16,13 @@ def handler(event, context):
         response = asgi_handler(event, context)
         return response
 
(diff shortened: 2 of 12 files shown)


api/main.py
```diff
@@ -15,6 +16,14 @@ def handler(event, context):
         response = asgi_handler(event, context)
         return response
 
+    elif "migrate" in event:
```
dj2, line 19: Won't this get picked up by the if on line 12 as this is an http method? Does this need to be handler inside mangum?

CalvinRodo: It shouldn't be as it should only be triggered by a custom invocation from us in the  github workflow here: 
https://github.com/cds-snc/scan-websites/blob/693346427ff145379ab71855b917328ef7fc8558/.github/workflows/build_and_push.yml#L75-L80

CalvinRodo: although maybe we should be more explicit then just checking to see if the event has the word migrate in it.

.github/workflows/build_and_push.yml
```diff
@@ -68,4 +68,13 @@ jobs:
         run: |
           aws lambda update-function-code \
             --function-name ${{ matrix.image }} \
-            --image-uri $REGISTRY/${{ matrix.image }}:latest
\ No newline at end of file
+            --image-uri $REGISTRY/${{ matrix.image }}:latest
+
+      - name: Migrate Database
```
maxneuvians, line 73: This would invoke on every run triggered by the matrix, which there may be multiple, we will need a `matrix.image == api` if in there

api/main.py
maxneuvians: I would just run a full comparison on the event

.github/workflows/build_and_push.yml
CalvinRodo: yep good catch

api/main.py
CalvinRodo: I'm now explicitly checking if event["task"] is migrate

.github/workflows/build_and_push.yml
CalvinRodo: fixed

api/main.py
dj2: I think the thing I'm missing is, where does the `httpMethod` matched above come from? Are we writting that into the event object or does AWS add it? Does calling with 'invoke' not do an HTTP call?

CalvinRodo: It come from API gateway, invoking is not an httpMethod, the event payload you get depends on how the lambda is being invoked, a DynamoDB Event Stream is going to be different from an SNS queue, to an API Gateway event, etc... 

https://docs.aws.amazon.com/whitepapers/latest/serverless-architectures-lambda/the-event-object.html

CalvinRodo: I'm trying to find a list of schemas and I'm coming up empty but did find this lambda python powertools library that has an event source data class you can take a look at. 

https://awslabs.github.io/aws-lambda-powertools-python/latest/utilities/data_classes/

CalvinRodo: Ooh maybe this is it:
https://docs.aws.amazon.com/lambda/latest/dg/lambda-services.html

dj2: So, this would fail if I `curl`d the endpoint? Only by doing the `invoke` will it work? (Finding the contents of those events is hard, heh)

CalvinRodo: So the endpoint is actually api gateway so you can't hit the lambda directly unless you do use the `aws` cli and call `lambda invoke`.

Curling it should always give an API Endpoint event with the httpMethod key in it.

dj2 (approved):
Code LGTM. I have more questions about the handler, but unrelated to this change.

PR #48, opened 2021-08-06
feat/crawler

maxneuvians:
This PR adds a crawler function to the API. Currently there is no way to invoke the crawler function from the API as we still need to implement the models. However, the crawler works and will take a URL and crawl it to a depth of 2, starting at 1. This is something we can change easily by setting a `API_CRAWLER_DEPTH` environment variable on the API.

Crawled URLs are dispatched and to a `pub_sub.dispatch` function which can then run some logic to what `scanners` it should pass them in. Currently the `dispatch` will send them to the axe-core scanner.

We also had to switch from an alpine container to a debian container because playwright does not have any architecture for alpine.


api/crawler/crawler.py
```diff
@@ -0,0 +1,70 @@
...
+        item = {}
+        item["id"] = self.id
+        item["url"] = response.url
+        item["depth"] = curr_depth
+        item["referer"] = response.meta.get("referer", "")
+        pub_sub.dispatch(item)
+
+        if curr_depth < self.max_depth:
+            for a in LinkExtractor(self.allowed_domains).extract_links(response):
+                yield response.follow(
```
dj2, line 37: What is the value passed to yield here, true/false (given the value on line 46)? What does false mean, stop crawl? Is that what we want? Does that mean that if a URL 404's we'll stop processing all other URLs?

api/pub_sub/pub_sub.py
```diff
@@ -0,0 +1,30 @@
...
+from logger import log
+
+
+def dispatch(payload):
+    # Add some logic to figure out what queues to dispatch to
+    # Currently only dispatches to axe-core
+    AXE_CORE_URLS_TOPIC = os.environ.get("AXE_CORE_URLS_TOPIC", False)
+
+    # Add a key that can be linked back to the ID of the payload
+    payload["key"] = str(uuid.uuid4())
```
dj2, line 16: How does `key` differ from `id`? Does id not need to be unique?

CalvinRodo (approved):
:taco::cat:

api/tests/pub_sub/test_pub_sub.py
```diff
@@ -0,0 +1,31 @@
...
+
+from pub_sub import pub_sub
+from unittest.mock import ANY, MagicMock, patch
+
+
+@patch("pub_sub.pub_sub.send")
+@patch.dict(os.environ, {"AXE_CORE_URLS_TOPIC": "topic"}, clear=True)
+def test_dispatch_adds_a_key_and_calls_send(mock_send):
+    pub_sub.dispatch({})
+    mock_send.assert_called_once_with("topic", {"key": ANY})
```
dj2, line 11: Does `ANY` also match `None`? Should this check that there is actually a value set?

api/crawler/crawler.py
maxneuvians: As far as I understand the `scrapy` implementation, `parse` ultimately yields something that can be processed in at Item pipeline (https://docs.scrapy.org/en/latest/topics/loaders.html). We don't need to use that so it just yields a `True`. It could as well just yield `item`.

api/tests/pub_sub/test_pub_sub.py
maxneuvians: In this case the key can't be none as it is set above: https://github.com/cds-snc/scan-websites/pull/48/files#diff-a729d59856eca0b0fd214e23c00398598e7d5b7c5f03e52a28723b6118da05f5R16. The `key` implementation is currently specific to the axe-core scanner and therefore not a concern of `send`. We can iterate once we figure out what the model looks like with the hierarchy of ids.

api/pub_sub/pub_sub.py
maxneuvians: I changed the code so that every URL gets a parent_id and an id, with the parent_id being linked to the original "job". This way we can keep the id concept the same and just link on itself in a table. I have also adjusted the scanner code in to look for `id` instead of `key`

dj2: What about if we went with something like `job_id`? So, we get the `url` digitial.canada.ca, we then generate a UUID `job_id` and all of the sub-urls get the same `job_id`?

Then we end up with something like, a job is the full request. A job can run over multiple URLs.

api/tests/pub_sub/test_pub_sub.py
dj2: I'm more concerned with, if I change the code in dispatch and mess it up so we don't assign the key, will this test fail? (Hence, does ANY match None?)

maxneuvians: Fixed to using `id` instead of `key`

api/pub_sub/pub_sub.py
maxneuvians: The friction is that we save axe reports and screenshots in an S3 bucket which require a unique key so you somehow have to pull out those ids. The `parent_id` now is more or less the `job_id` - so we might as well rename it to that :p

api/tests/pub_sub/test_pub_sub.py
maxneuvians: Surprisingly hard in python, there seem to be no existing type checkers for `assert_called_once_with`

dj2 (approved):

PR #51, opened 2021-08-10
feat: add S3 bucket for axe core reports and SNS queue for axe-core-urls

maxneuvians:
This PR adds an S3 bucket to the infrastructure to collect axe-core reports. It also adds a SNS topic that will send URLs to be scanned by axe-core.

### DIFF
--- a/terragrunt/aws/api/kms.tf
+++ b/terragrunt/aws/api/kms.tf
@@ -1,4 +1,4 @@
-  
+
 data "aws_caller_identity" "current" {}
 
 resource "aws_kms_key" "scan-websites" {
--- a/terragrunt/aws/api/kms.tf
+++ b/terragrunt/aws/api/kms.tf
@@ -1,0 +1,53 @@
+  
+data "aws_caller_identity" "current" {}
+
+resource "aws_kms_key" "scan-websites" {
+  description         = "KMS Key"
+  enable_key_rotation = true
+
+  policy = <<EOF
+{
+  "Version" : "2012-10-17",
+  "Id" : "key-default-1",
+  "Statement" : [ {
+      "Sid" : "Enable IAM User Permissions",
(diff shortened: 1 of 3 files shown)


terragrunt/aws/api/kms.tf
```diff
@@ -0,0 +1,53 @@
...
+resource "aws_kms_key" "scan-websites" {
+  description         = "KMS Key"
+  enable_key_rotation = true
+
+  policy = <<EOF
+{
+  "Version" : "2012-10-17",
+  "Id" : "key-default-1",
+  "Statement" : [ {
+      "Sid" : "Enable IAM User Permissions",
```
CalvinRodo, line 13: Instead of a HereDoc would be nice if this was a aws_policy_document

CalvinRodo: Alternatively we could use `jsonencode()`

terragrunt/aws/api/s3.tf
```diff
@@ -0,0 +1,14 @@
+resource "aws_s3_bucket" "axe-core-report-data" {
+  bucket = "${var.product_name}-${var.env}-axe-core-report-data"
```
CalvinRodo, line 2: Should we append a random key to this?

CalvinRodo (changes requested):
See my comments

terragrunt/aws/api/s3.tf
maxneuvians: Should be unique enough

terragrunt/aws/api/kms.tf
maxneuvians: using a `aws_iam_policy_document` now

CalvinRodo (approved):
:taco::cat:

PR #52, opened 2021-08-10
feat: infrastructure for zap driven website scans

mohamed-cds:
AWS infrastructure for OWASP Zap driven website scanning. When complete the zap process will read URL's from a SNS queue and start a ECS task to scan all the URL's using OWASP Zap


scanners/zap-lambda/src/common/foreach.ts
```diff
@@ -0,0 +1,9 @@
+// Helper function to run asynchronous foreach loop.
```
dj2, line 1: Instead of duplicating these, can we move the common folder from axe-core up to share with the zap-lambda?

mohamed-cds: refactored. Thanks!

mohamed-cds: i'll take that back. Needs more work so that it works when building the container by copying in the files at buildtime

dj2: Ok, I can see if I can fixup the duplication after this lands.

mohamed-cds: Thanks! I tried to find a solution but ran into build context issues when doing docker build and using common makefile + common/ library.

scanners/owasp-zap/Dockerfile
```diff
@@ -0,0 +1,81 @@
...
+# Create function directory
+RUN mkdir -p ${FUNCTION_DIR}
+
+# Copy handler function
+COPY . ${FUNCTION_DIR}
+
+WORKDIR ${FUNCTION_DIR}
+
+# Optional – Install the function's dependencies
+ENV PUPPETEER_SKIP_CHROMIUM_DOWNLOAD=true
```
dj2, line 47: Does zap use puppeteer? Is this needed?

scanners/owasp-zap/src/impl.ts
```diff
@@ -0,0 +1,47 @@
+import { ECS } from "aws-sdk";
+import { asyncForEach } from "./common/foreach";
+import { Record } from "./common/record";
+
+const ecs = new ECS({ region: "ca-central-1" });
```
dj2, line 5: The initial idea with the `impl` files is that they would know nothing about AWS. So, if you want an `ecs` instanced you'd pass it in from the index file and we'd define an `interface` which just has the methods you care about. That way we can inject something that isn't AWS in testing (or use outside AWS by implementing the interface)

scanners/owasp-zap/src/impl.ts
```diff
@@ -0,0 +1,47 @@
...
+export async function Impl(
+  records: Record[],
+): Promise<boolean> {
+  try {
+    await asyncForEach(records, async (record: Record) => {
+       const params = {
+        launchType: "FARGATE",
+        cluster: process.env.CLUSTER,
+        taskDefinition: process.env.TASK_DEF_ARN,
+        overrides: {containerOverrides: [{name: "owasp_zap",environment: [{name: "SCAN_URL", value: record.payload.url}]}]},
```
dj2, line 16: In the axe-core side we defined a JOB_Id which is the same for all of the URLs from a given request. Do we want something similar here so we can tie all of the things that are run from a given request together?

scanners/owasp-zap/src/impl.ts
```diff
@@ -0,0 +1,47 @@
...
+  } catch (error) {
+    console.log(error);
+    return false;
+  }
+  return true;
+}
+
+export async function convertEventToRecords(
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any, @typescript-eslint/explicit-module-boundary-types
+  event: any,
```
dj2, line 31: This `any` I think, could be an `SNSEvent` to remove the lint disable.

scanners/owasp-zap/src/impl.ts
```diff
@@ -0,0 +1,47 @@
...
+): Promise<Record[]> {
+  const records: Record[] = [];
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  await asyncForEach(event.Records, async (record: any) => {
+    // Parse the correct message body, SNS or S3
```
dj2, line 36: Is this true? I think we're only dealing with SNS here?

scanners/owasp-zap/src/impl.ts
```diff
@@ -0,0 +1,47 @@
...
+    // eslint-disable-next-line no-prototype-builtins
+    records.push({
+      payload: JSON.parse(record.Sns.Message),
+      html: "",
+    });
+  });
+  return records;
+}
+
+export const isStringEmptyUndefinedOrNull = (str: string): boolean =>
```
dj2, line 46: Unused except for the tests, remove?

mohamed-cds: removed

scanners/owasp-zap/src/impl.ts
mohamed-cds: Updated to SNSEventRecord and updated the comment

scanners/owasp-zap/src/impl.ts
mohamed-cds: Updated to SNSEvent

scanners/owasp-zap/Dockerfile
mohamed-cds: copy paste fail. Removed

scanners/owasp-zap/src/impl.ts
mohamed-cds: Most definitely. I'd like to send all URL's that were crawled from the same source to ZAP for the security scan. I'll look at how axe-core does it.

mohamed-cds: i'll create an issue to migrate this over to JOB_Id once i get the basic scanning working here.

scanners/owasp-zap/src/impl.ts
mohamed-cds: I'll do the same, new commit incoming.

patheard (approved):

PR #53, opened 2021-08-11
feat: add api to docker compose network

CalvinRodo:
# Summary | Résumé

- clean up the docker compose file
- add scripts to execute the lambda endpoints in the api service
- download a pinned version of the aws-lambda-rie Executable
- add some logging of env variables to the api
- add shellcheck as a ci check

### DIFF
--- a/.devcontainer/docker-compose.yml
+++ b/.devcontainer/docker-compose.yml
@@ -1,5 +1,9 @@
 version: '3'
 
+x-db-connection-strings: &db-connection-strings
+  SQLALCHEMY_DATABASE_URI: postgresql://postgres:postgres@db/scan-websites
+  SQLALCHEMY_DATABASE_TEST_URI: postgresql://postgres:postgres@test-db/scan-websites
+
 services:
   app:
     build: 
@@ -18,15 +22,13 @@ services:
     volumes:
       - ..:/workspace:cached   
     command: sleep infinity
-    ports: 
-      - 8000:8000
-    links: 
-      - db
-      - test-db
-    environment: 
-      SQLALCHEMY_DATABASE_URI: postgresql://postgres:postgres@db/scan-websites
-      SQLALCHEMY_DATABASE_TEST_URI: postgresql://postgres:postgres@test-db/scan-websites
-
+    environment:
+      <<: *db-connection-strings
+  api:
+    build:
+      context: ../api/
+    environment:
+      <<: *db-connection-strings
   db:
     image: postgres:11.2
     volumes:
--- a/api/Dockerfile
+++ b/api/Dockerfile
@@ -65,8 +65,6 @@ ENV GIT_SHA=$git_sha
 
 # Install lambda runtime interactive environment
 ARG RIE_VERSION=1.1
-# RUN wget https://github.com/aws/aws-lambda-runtime-interface-emulator/releases/latest/download/aws-lambda-rie \
-#     && mv aws-lambda-rie /usr/bin/aws-lambda-rie
 RUN wget -O aws-lambda-rie https://github.com/aws/aws-lambda-runtime-interface-emulator/releases/download/${RIE_VERSION}/aws-lambda-rie \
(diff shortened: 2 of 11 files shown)


.devcontainer/docker-compose.yml
```diff
@@ -18,15 +22,13 @@ services:
...
-      - 8000:8000
-    links: 
-      - db
-      - test-db
-    environment: 
-      SQLALCHEMY_DATABASE_URI: postgresql://postgres:postgres@db/scan-websites
-      SQLALCHEMY_DATABASE_TEST_URI: postgresql://postgres:postgres@test-db/scan-websites
-
+    environment:
+      <<: *db-connection-strings
```
dj2, line 26: So the `x-db-connection-strings: &db-connection-strings` line declares the `db-connection-strings` key? Is that how this is read?

api/Dockerfile
```diff
@@ -63,8 +63,12 @@ COPY --from=build-image ${FUNCTION_DIR} ${FUNCTION_DIR}
 ARG git_sha 
 ENV GIT_SHA=$git_sha
 
-# (Optional) Add Lambda Runtime Interface Emulator and use a script in the ENTRYPOINT for simpler local runs
-ADD https://github.com/aws/aws-lambda-runtime-interface-emulator/releases/latest/download/aws-lambda-rie /usr/bin/aws-lambda-rie
+# Install lambda runtime interactive environment
+ARG RIE_VERSION=1.1
+# RUN wget https://github.com/aws/aws-lambda-runtime-interface-emulator/releases/latest/download/aws-lambda-rie \
```
dj2, line 68: Can we remove this if it isn't needed?

bin/api_version.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+echo hitting api version endpoint
+curl "http://api:8080/2015-03-31/functions/function/invocations" -d '{
```
dj2, line 4: Are these to be run from inside a container on AWS?

scanners/axe-core/Dockerfile
```diff
@@ -76,7 +76,9 @@ ENV GIT_SHA=$git_sha
 ENV PUPPETEER_EXECUTABLE_PATH=/usr/bin/chromium-browser
 
 # (Optional) Add Lambda Runtime Interface Emulator and use a script in the ENTRYPOINT for simpler local runs
-ADD https://github.com/aws/aws-lambda-runtime-interface-emulator/releases/latest/download/aws-lambda-rie /usr/bin/aws-lambda-rie
+# Install lambda runtime interactive environment
+RUN curl -Lo aws-lambda-rie https://github.com/aws/aws-lambda-runtime-interface-emulator/releases/download/${RIE_VERSION}/aws-lambda-rie \
```
dj2, line 80: Does `RIE_VERSION` get inherited from the other dockerfile? Otherwise, I don't think it's set in here.

.devcontainer/docker-compose.yml
CalvinRodo: These are extension fields. 

https://docs.docker.com/compose/compose-file/compose-file-v3/#extension-fields

So you are correct it declares some common fields and appends them to the `environment` fields.

api/Dockerfile
CalvinRodo: Ah yes the code was just for testing

scanners/axe-core/Dockerfile
CalvinRodo: No it does not...

bin/api_version.sh
CalvinRodo: yes they are to be run inside the dev container. 

we can run them outside of the dev container but we would have to add the port mappings back

CalvinRodo: and you would still need to have the dev container docker compose running.

dj2 (approved):

PR #54, opened 2021-08-12
Feat/axe core ecr

maxneuvians:
The PR adds an ECR for the axe-core-scanner image. It also scaffold the scanners/axe-core module in terragrunt.

### DIFF
--- a/terragrunt/aws/scanners/axe-core/ecr.tf
+++ b/terragrunt/aws/scanners/axe-core/ecr.tf
@@ -1,0 +1,10 @@
+resource "aws_ecr_repository" "scanners-axe-core" {
+  # checkov:skip=CKV_AWS_51:The :latest tag is used in Staging
+
+  name                 = "${var.product_name}/scanners/axe-core"
+  image_tag_mutability = "MUTABLE"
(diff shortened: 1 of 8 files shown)


mohamed-cds (approved):
nice 🚀 ✅ . I'm going to steal your naming convention 😄

terragrunt/aws/scanners/axe-core/ecr.tf
```diff
@@ -0,0 +1,10 @@
+resource "aws_ecr_repository" "scanners-axe-core" {
+  # checkov:skip=CKV_AWS_51:The :latest tag is used in Staging
+
+  name                 = "${var.product_name}/scanners/axe-core"
+  image_tag_mutability = "MUTABLE"
```
CalvinRodo, line 5: Is the plan to have the lambda pull the latest tag all the time?

If not we should remove mutability

maxneuvians: Yes otherwise you always have to push a SHA which you then have to make sure the TF apply runs after the docker build.

PR #56, opened 2021-08-12
feat: add axe-core scanner lambda

maxneuvians:
Adds the lambda to the axe-core scanner. The lambda is triggered through messages in an SNS queue and will write reports to an S3 bucket and also take a screenshot of the URL it visits.

### DIFF
--- a/terragrunt/aws/scanners/axe-core/iam.tf
+++ b/terragrunt/aws/scanners/axe-core/iam.tf
@@ -1,0 +1,85 @@
+data "aws_iam_policy_document" "service_principal" {
+  statement {
+    effect = "Allow"
+
+    actions = ["sts:AssumeRole"]
+
+    principals {
+      type        = "Service"
+      identifiers = ["lambda.amazonaws.com"]
+    }
+  }
+}
+
+resource "aws_iam_role" "scanners-axe-core" {
+  name               = "${var.product_name}-scanners-axe-core"
+  assume_role_policy = data.aws_iam_policy_document.service_principal.json
+}
+
+data "aws_iam_policy" "lambda_insights" {
+  name = "CloudWatchLambdaInsightsExecutionRolePolicy"
+}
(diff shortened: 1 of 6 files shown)


mohamed-cds (approved):
coming along real well! 🚀 ✅

terragrunt/aws/scanners/axe-core/iam.tf
```diff
@@ -0,0 +1,85 @@
...
+
+  statement {
+
+    effect = "Allow"
+
+    actions = [
+      "s3:ListBucket",
+      "s3:ListBucketVersions",
+      "s3:GetBucketLocation",
+      "s3:Get*",
```
dj2, line 66: I'm not sure how to read this, but do we have to restrict access to these screenshots? If we allow running with credentials there maybe stuff in there that shouldn't just be public.

mohamed-cds: By default the s3 objects in the bucket will only be accessible to users with either console or programmatic access to the AWS account. The above statement is granting the scanners-axe-core iam role the ability to interact with the s3 bucket, this role will be used by the axe-core lambda to save reports/screens. Users with credentialed access to scan websites will not inherit these permissions.

dj2: In the previous iteration of this the screenshots were displayed on the website. I'm assuming that's going to be the case here as well. How do we lock down the screenshots of credentialed sites? Will we be adding new rules in the future?

mohamed-cds: one option is to route the image get requests through an api that would retrieve the images from the s3 bucket. This way we don't have to make the bucket publicly accessible

dj2: +1, that's a great idea.

PR #61, opened 2021-08-12
feat: adds the S3 report trigger to the API and adds sample handling

maxneuvians:
This PR adds a trigger to the API lambda to be invoked if an S3 file is added to the axe-core-report-data bucket. The API will invoke, download the file and produce a sample log output. Ideally at this point the API would extract the summary data from the report, write a summary into the database and do whatever else is need for reporting.

### DIFF
--- a/api/main.py
+++ b/api/main.py
@@ -27,6 +27,8 @@ def handler(event, context):
         for record in event.get("Records", []):
             if "s3" in record:
                 storage.get_object(record)
+            else:
+                log.warning(f"Handler recieved unrecognised record: {record}")
         return "Success"
 
     elif event.get("task", "") == "migrate":
--- a/api/main.py
+++ b/api/main.py
@@ -2,6 +2,7 @@ from mangum import Mangum
 from api_gateway import api
 from logger import log
 from database.migrate import migrate_head
+from storage import storage
 import os
 
 app = api.app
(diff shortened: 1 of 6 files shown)


api/main.py
```diff
@@ -22,6 +23,12 @@ def handler(event, context):
         response = asgi_handler(event, context)
         return response
 
+    elif "Records" in event:
+        for record in event.get("Records", []):
+            if "s3" in record:
+                storage.get_object(record)
+        return "Success"
```
dj2, line 30: If we receive an SNS event, I think it also has Records but wouldn't have the s3 key, this would still return Success in that case when we didn't process any events.

api/storage/storage.py
```diff
@@ -0,0 +1,21 @@
+import boto3
+from logger import log
+
+
+def get_object(record):
```
dj2, line 5: This probably needs a different name depending on what that comment expands into, it soon won't be just getting an object, or is the plan to pass in a lambda to do the work so storage stays generic?

api/main.py
maxneuvians: You are right - better to be explicit than fail silently

api/storage/storage.py
maxneuvians: I figured all S3 objects need to be loaded and then from there you can pass them to different functions based on their content/bucket name

dj2: I that case, this should just load/return the content. Let the caller process as they want. Split the caller from the api class out into it's own method that calls storage and then does processing.

Keeps this simpler.

api/main.py
maxneuvians: It now logs that it did not know how to handle an event

api/storage/storage.py
maxneuvians: it just returns the data now

api/main.py
dj2: nit: s/recieved/received

dj2 (approved):

PR #62, opened 2021-08-13
Additions and updates to makefiles

dj2:
# Summary | Résumé

This PR adds more makefile targets, moves terragrunt commands into the
makefiles and makes sure the terraform commands are run during the
lint/formatting passes.


.github/workflows/tf_apply.yml
```diff
@@ -49,20 +49,7 @@ jobs:
...
-        working-directory: terragrunt/env/scanners/axe-core
-        run: |
-          terragrunt apply --terragrunt-non-interactive -auto-approve
-
-      - name: Apply scanners/owasp-zap
-        if: ${{ steps.filter.outputs.scanners-owasp-zap == 'true' }}
-        working-directory: terragrunt/env/scanners/owasp-zap
-        run: |
-          terragrunt apply --terragrunt-non-interactive -auto-approve          
+          make terragrunt-apply
```
patheard, line 55: The impact here is that:
* `terragrunt apply` will always run for all modules (previously it was only run for a module if there were changes); and
* it will only run if there are changes to the API module specifically.

dj2:
Closing this. Was pointed out that there are ordering dependencies in the terragrunt apply that this won't honour. The `api` module, if changed, must run before the other modules. Going to split some of the other changes for lint and checkov out to a new PR.

PR #70, opened 2021-08-13
Update some naming.

dj2:
# Summary | Résumé

The python-test action is renamed to test as it runs more then just
python. The dev container is renamed to `Scan Websites` to make it
clearer which one you're working with.


CalvinRodo (approved):
:taco::cat:

PR #74, opened 2021-08-13
feat: user model

### LINKED ISSUE #69: Add User model
Add `User` model based on the proposal in #9

maxneuvians:
Closes #69. Adds a `user` model based on the proposal in #9.

### DIFF
--- a/api/db_migrations/versions/b9ab107cd56a_create_users_table.py
+++ b/api/db_migrations/versions/b9ab107cd56a_create_users_table.py
@@ -1,0 +1,41 @@
+"""create_users_table
+
+Revision ID: b9ab107cd56a
+Revises: 06b310a6251e
+Create Date: 2021-08-13 16:10:56.221141
+
+"""
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+
+# revision identifiers, used by Alembic.
+revision = "b9ab107cd56a"
+down_revision = "06b310a6251e"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.create_table(
+        "users",
+        sa.Column("id", postgresql.UUID(as_uuid=True), primary_key=True),
+        sa.Column("organisation_id", postgresql.UUID(as_uuid=True), nullable=False),
+        sa.Column("name", sa.String(), nullable=False),
+        sa.Column("email_address", sa.String(length=255), nullable=False, unique=True),
+        sa.Column("password_hash", sa.String(length=255), nullable=False),
+        sa.Column(
+            "access_token", postgresql.UUID(as_uuid=True), nullable=False, unique=True
+        ),
+        sa.Column("created_at", sa.DateTime, default=sa.func.now()),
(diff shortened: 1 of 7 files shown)


api/db_migrations/versions/b9ab107cd56a_create_users_table.py
```diff
@@ -0,0 +1,41 @@
...
+def upgrade():
+    op.create_table(
+        "users",
+        sa.Column("id", postgresql.UUID(as_uuid=True), primary_key=True),
+        sa.Column("organisation_id", postgresql.UUID(as_uuid=True), nullable=False),
+        sa.Column("name", sa.String(), nullable=False),
+        sa.Column("email_address", sa.String(length=255), nullable=False, unique=True),
+        sa.Column("password_hash", sa.String(length=255), nullable=False),
+        sa.Column(
+            "access_token", postgresql.UUID(as_uuid=True), nullable=False, unique=True
```
dj2, line 29: Probably want an index on this one.

api/db_migrations/versions/b9ab107cd56a_create_users_table.py
```diff
@@ -0,0 +1,41 @@
...
+depends_on = None
+
+
+def upgrade():
+    op.create_table(
+        "users",
+        sa.Column("id", postgresql.UUID(as_uuid=True), primary_key=True),
+        sa.Column("organisation_id", postgresql.UUID(as_uuid=True), nullable=False),
+        sa.Column("name", sa.String(), nullable=False),
+        sa.Column("email_address", sa.String(length=255), nullable=False, unique=True),
```
dj2, line 26: index? (I'm assuming login is email/password)

api/tests/models/test_User.py
```diff
@@ -0,0 +1,106 @@
...
+        password="password",
+        organisation=organisation_fixture,
+    )
+    session.add(user)
+    with pytest.raises(IntegrityError):
+        session.commit()
+    session.rollback()
+
+
+def test_user_empty_email_address_fails(organisation_fixture, session):
```
dj2, line 73: Add a test_user_duplicate_email_address test

api/db_migrations/versions/b9ab107cd56a_create_users_table.py
maxneuvians: I thought unique columns don't need to be indexed because unique creates an implicit index? Happy to add one if this is not correct.

api/db_migrations/versions/b9ab107cd56a_create_users_table.py
maxneuvians: See above

api/tests/models/test_User.py
maxneuvians: Good call! Fixed

api/db_migrations/versions/b9ab107cd56a_create_users_table.py
dj2: I don't know alchemy, so maybe it does? We should verify that the index gets created though as we'll probably want to lookup by access_token a lot.

dj2 (approved):
We can always fixup the index later if it doesn't get created.

PR #75, opened 2021-08-16
Add template tsconfig and makefile for scanners.

dj2:
# Summary | Résumé

This PR pulls the common makefile and tsconfig information up a level
and updates the two scanners to share the configuration.


dj2:
@mohamed-cds this adds owasp-zap into the RESOURCES so all the make commands run against it. Was it left out on purpose? Seemed like an oversight?

mohamed-cds:
> @mohamed-cds this adds owasp-zap into the RESOURCES so all the make commands run against it. Was it left out on purpose? Seemed like an oversight?

wasn't intentional 👍 I think it was accidently dropped during my refactor

.github/workflows/scripts/build_container.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+
```
patheard, line 2: Might be worth enabling [bash strict mode](http://redsymbol.net/articles/unofficial-bash-strict-mode/).  It has saved me more than once:
```sh
#!/bin/bash
set -euo pipefail
IFS=$'\n\t'
```

dj2: Cool. Done.

.github/workflows/scripts/build_container.sh
```diff
@@ -0,0 +1,29 @@
...
+IMAGE=$3
+
+EXTRA_ARGS=()
+if [[ "$IMAGE" == "scanners/axe-core" ]];
+then
+  EXTRA_ARGS+=(--build-arg CHROMIUM=with-chromium)
+fi
+
+DOCKER_FILE="Dockerfile"
+if [[ $IMAGE == scanners/* ]];
```
patheard, line 16: nit - quotes around var and string:
```sh
if [[ "$IMAGE" == "scanners/*" ]];
```

patheard (commented):
Minor quote comment, but for the good idea vs. bad idea debate, I'm all for things that get rid of duplication.

.github/workflows/scripts/build_container.sh
dj2: The lack of quotes on scanners/* was intentional as I want it to glob, not strictly match.

patheard: Ah, my bad

patheard (approved):

PR #77, opened 2021-08-16
Feat/zap ecs container

mohamed-cds:
ECS container that will initiate the zap scan and save the report to S3 for future processing.

ECS is being used instead of a lambda due to vulnerability scans typically exceeding the lambda maximum execution time of 15 minutes.


CalvinRodo (approved):
:taco::cat

runners/owasp-zap/entrypoint.sh
```diff
@@ -0,0 +1,51 @@
...
+  HOST_IP="172.17.0.1"
+else
+  # Use ECS host IP
+  HOST_IP=$(curl -s "$ECS_CONTAINER_METADATA_URI" | jq -r '.Networks[].IPv4Addresses[0]')
+fi
+
+echo "Host ip: $HOST_IP and Port:${ZAP_PORT}"
+
+# Wait for ZAP proxy to init
+while ! eval "$(curl -sSf "$HOST_IP":"${ZAP_PORT}" > /dev/null 2>&1)"
```
patheard, line 15: It might make sense to have an exit condition from this in case the ZAP proxy never inits (e.g. exit after `x` number of failed waits).

patheard (commented):
Had a comment about the `entrypoint.sh` while loop waiting for ZAP proxy init.

runners/owasp-zap/entrypoint.sh
```diff
@@ -0,0 +1,51 @@
...
+do
+	echo "Waiting for proxy to start..."
+	sleep 3
+done
+sleep 3
+
+date=$(date +\"%Y-%m-%dT%H:%M:%S\")
```
dj2, line 22: Do we expect second granularity to be enough for the filenames? What happens if 2 PRs get submitted at the same time, or is the runner serial?

runners/owasp-zap/entrypoint.sh
```diff
@@ -0,0 +1,51 @@
...
+fDate=$(echo "$date" | sed -e 's/[^A-Za-z0-9._-]/_/g')
+# Convert URL into a valid filename for the report
+FILENAME=$(echo "$SCAN_URL" | sed -e 's/[^A-Za-z0-9._-]/_/g')-$fDate
+
+zap-cli --port "${ZAP_PORT}" --zap-url "http://$HOST_IP" exclude "^.*(logout|log-out|signout|sign-out).*$"
+zap-cli --port "${ZAP_PORT}" --zap-url "http://$HOST_IP" exclude "^.*\.(css|gif|jpe?g|tiff|png|webp|bmp|ico|svg|js|jsx)$"
```
dj2, line 28: pdf? What about various video formats or does it exclude those already?

runners/owasp-zap/entrypoint.sh
```diff
@@ -0,0 +1,51 @@
...
+zap-cli --port "${ZAP_PORT}" --zap-url "http://$HOST_IP" open-url "${SCAN_URL}"
+zap-cli --port "${ZAP_PORT}" --zap-url "http://$HOST_IP" spider "${SCAN_URL}"
+zap-cli --port "${ZAP_PORT}" --zap-url "http://$HOST_IP" ajax-spider "${SCAN_URL}"
+
+# Timeout scan after 1 hour to prevent running indefinately if the OWASP ZAP container crashes
+timeout 1h zap-cli --port "${ZAP_PORT}" --zap-url "http://$HOST_IP" active-scan --recursive "${SCAN_URL}"
```
dj2, line 34: 1 hour seems really long, how long do these take on average?

runners/owasp-zap/entrypoint.sh
```diff
@@ -0,0 +1,51 @@
...
+
+high_alerts=$( curl "http://$HOST_IP:${ZAP_PORT}/JSON/alert/view/alertsSummary/?baseurl=${SCAN_URL}" | jq -r '.alertsSummary.High')
+
+echo "high alerts are $high_alerts"
+
+curl "http://$HOST_IP:${ZAP_PORT}/OTHER/core/other/jsonreport/" | jq . > zap-scan-results.json
```
dj2, line 40: Does the runner run serially? If not, could there be a race for zap-scan-results.json?

runners/owasp-zap/entrypoint.sh
mohamed-cds: For small apps it'll take around 15-30 minutes. For larger apps it can take a few hours. This will also vary depending on how fast the webapp responds to requests

runners/owasp-zap/entrypoint.sh
mohamed-cds: each scan starts up a dedicated instance of owasp zap proxy, so there will never be any collisions.

runners/owasp-zap/entrypoint.sh
```diff
@@ -0,0 +1,58 @@
...
+  fi
+done
+sleep 3
+
+date=$(date +\"%Y-%m-%dT%H:%M:%S\")
+fDate=$(echo "$date" | sed -e 's/[^A-Za-z0-9._-]/_/g')
+# Convert URL into a valid filename for the report
+FILENAME=$(echo "$SCAN_URL" | sed -e 's/[^A-Za-z0-9._-]/_/g')-$fDate
+
+zap-cli --port "${ZAP_PORT}" --zap-url "http://$HOST_IP" exclude "^.*(logout|log-out|signout|sign-out).*$"
```
patheard, line 34: Dan's comment got me thinking - since we're excluding English logout paths, it would make sense to also exclude French logout paths as well:
```regex
^.*(logout|log-out|signout|sign-out|deconnecter).*$
```

runners/owasp-zap/entrypoint.sh
mohamed-cds: i started with the common list of exclusions, added pdf. video formats are rarely seen when scanning websites these days since they're usually behind some sort of cdn with a different domain than the app being scanned.

patheard (commented):
Comment about French logout path exclusion.

runners/owasp-zap/entrypoint.sh
mohamed-cds: Good point. Nanosecond added

runners/owasp-zap/entrypoint.sh
mohamed-cds: Good point! Added an exit condition

runners/owasp-zap/entrypoint.sh
mohamed-cds: 100%. Added :)

runners/owasp-zap/entrypoint.sh
```diff
@@ -0,0 +1,58 @@
...
+  fi
+done
+sleep 3
+
+date=$(date +\"%Y-%m-%dT%H:%M:%S:%N\")
+fDate=$(echo "$date" | sed -e 's/[^A-Za-z0-9._-]/_/g')
+# Convert URL into a valid filename for the report
+FILENAME=$(echo "$SCAN_URL" | sed -e 's/[^A-Za-z0-9._-]/_/g')-$fDate
+
+zap-cli --port "${ZAP_PORT}" --zap-url "http://$HOST_IP" exclude "^.*(logout|log-out|signout|sign-out|deconnecter).*$"
```
dj2, line 34: This might be a little lax. It would match '/blog/how-we-implemented-logout' or '/blog/our-sign-out-system', etc.

patheard: Tighten with slashes as a first step?
```regex
^.*/(logout|log-out|signout|sign-out|deconnecter)/.*
```

patheard (approved):

runners/owasp-zap/entrypoint.sh
dj2: Maybe `^.*/(logout|log-out|signout|sign-out|deconnecter)/?$` so the trailing slash is optional and logout is end of string?

mohamed-cds: 100%. This is why we'll always need to spend time manually testing since automation has it's limits.

When testing manually you scope exclusions to full URL paths, but with automation we have to use a heavier hand since the impact of accidentally crawling logout has a greater impact than not scanning URL's that contain the same keywords.

I'll add an issue to look into providing URL exclusions during API invocation for scanning. Once we have user input on what not to scan we can remove this exclusion.

mohamed-cds: Suggested regex implemented and created #79

dj2 (approved):

PR #81, opened 2021-08-17
chore: reduce zap image size

mohamed-cds:
Pip install of awscli was resulting in the runner-owasp-zap image being > 2 GB. Downloading the zip from aws and copying it to the bin path allows the image sizes to be reduced significantly.

### DIFF
--- a/runners/owasp-zap/Dockerfile
+++ b/runners/owasp-zap/Dockerfile
@@ -6,7 +6,14 @@ RUN apk update \
     && apk upgrade \
     && apk add --update curl bash jq
 
-RUN pip install --upgrade zapcli awscli
+RUN pip install --upgrade zapcli
+
+# Install AWS CLI
+ARG AWS_CLI_VERSION="2.2.29"
+RUN curl -Lo awscliv2.zip "https://awscli.amazonaws.com/awscli-exe-linux-x86_64-${AWS_CLI_VERSION}.zip" \
+    && unzip awscliv2.zip \
+    && ./aws/install -i /usr/local/aws-cli -b /usr/local/bin
+
 COPY entrypoint.sh /entrypoint.sh
 
 # Launch OWASP scan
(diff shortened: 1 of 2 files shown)


CalvinRodo (approved):
:taco::cat:

runners/owasp-zap/Dockerfile
```diff
@@ -6,7 +6,14 @@ RUN apk update \
...
 
-RUN pip install --upgrade zapcli awscli
+RUN pip install --upgrade zapcli
+
+# Install AWS CLI
+ARG AWS_CLI_VERSION="2.2.29"
+RUN curl -Lo awscliv2.zip "https://awscli.amazonaws.com/awscli-exe-linux-x86_64-${AWS_CLI_VERSION}.zip" \
+    && unzip awscliv2.zip \
+    && ./aws/install -i /usr/local/aws-cli -b /usr/local/bin
+
```
patheard, line 16: We should be able to reduce image size by deleting the downloaded `.zip` at the end of the RUN command:
```
RUN curl -Lo awscliv2.zip "https://awscli.amazonaws.com/awscli-exe-linux-x86_64-${AWS_CLI_VERSION}.zip" \
    && unzip awscliv2.zip \
    && ./aws/install -i /usr/local/aws-cli -b /usr/local/bin \
    && rm awscliv2.zip
```

runners/owasp-zap/Dockerfile
```diff
@@ -6,7 +6,14 @@ RUN apk update \
     && apk upgrade \
     && apk add --update curl bash jq
 
-RUN pip install --upgrade zapcli awscli
+RUN pip install --upgrade zapcli
+
+# Install AWS CLI
+ARG AWS_CLI_VERSION="2.2.29"
+RUN curl -Lo awscliv2.zip "https://awscli.amazonaws.com/awscli-exe-linux-x86_64-${AWS_CLI_VERSION}.zip" \
```
patheard, line 13: Do we want to do the same signature comparison that's happening in the devcontainer for this?
https://github.com/cds-snc/scan-websites/blob/b34299cdcadee1f03eed14dfb2a5c4059c61aea4/.devcontainer/Dockerfile#L50-L57

patheard (commented):
Comments about deleting zip file and signature validation.

PR #84, opened 2021-08-17
feat: replace s3 with s3 modules in API module

CalvinRodo:
# Summary | Résumé

- Replace s3 resources with S3 modules from the TF Modules repo 
- Add a log bucket to store access logs on our buckets
- scanner/owasp-zap has a dependency on the log bucket so will be done in a followup PR


patheard (approved):
Looks good, but I'm wondering if the `tf apply` will fail since both buckets being destroyed have `force_destroy = false`.

mohamed-cds (approved):
deny_insecure_transport & name-prefix to allow us to easily pivot to supporting multiple environments are nice additions! Good stuff 🚀 ✅

PR #86, opened 2021-08-18
fix: bug fixes related to zap scanning startup

mohamed-cds:
Zap scan lambda will read from the SNS topic and launch a new ECS task that will scan the URL provided and store the results in the owasp-zap report bucket. Reading the report from the bucket and storing it in the database and optionally aws security hub will be done seperately.

- Install glibc in alpine container so that awscliv2 installs correctly
- Fixes for S3 iam permissions
- Fixed container definition to reference correct docker image


runners/owasp-zap/Dockerfile
```diff
@@ -2,18 +2,36 @@ FROM python:3.8-alpine
...
 WORKDIR /scan
 
+ENV GLIBC_VER=2.31-r0
+ARG AWS_CLI_VERSION="2.2.29"
+
+# install glibc compatibility for alpine (Needed for awscliv2)
 RUN apk update \
     && apk upgrade \
-    && apk add --update curl bash jq
+    && apk add --update curl binutils bash jq \
```
patheard, line 11: You should be able to switch the `--update` to `--no-cache` which does an implicit `apk update` first and doesn't use a local index.

runners/owasp-zap/Dockerfile
```diff
@@ -2,18 +2,36 @@ FROM python:3.8-alpine
...
+        awscliv2.zip \
+        aws \
+        /usr/local/aws-cli/v2/*/dist/aws_completer \
+        /usr/local/aws-cli/v2/*/dist/awscli/data/ac.index \
+        /usr/local/aws-cli/v2/*/dist/awscli/examples \
+    && apk --no-cache del \
+        binutils \
+    && rm glibc-${GLIBC_VER}.apk \
+    && rm glibc-bin-${GLIBC_VER}.apk \
+    && rm -rf /var/cache/apk/*
```
patheard, line 31: Once all the `apk add` commands are using `--no-cache`, you can get rid of this line.

runners/owasp-zap/Dockerfile
```diff
@@ -2,18 +2,36 @@ FROM python:3.8-alpine
 
 WORKDIR /scan
 
+ENV GLIBC_VER=2.31-r0
```
patheard, line 5: If you only need `GLIBC_VER` when building the image, you can switch this to an ARG:
```dockerfile
ARG GLIBC_VER="2.31-r0"
```

patheard (commented):
Looks good!  Had a few comments for the `Dockerfile` but nothing major.

runners/owasp-zap/Dockerfile
mohamed-cds: changed, thanks!

runners/owasp-zap/Dockerfile
mohamed-cds: changed, thanks!

runners/owasp-zap/Dockerfile
mohamed-cds: changed, thanks!

patheard (approved):

runners/owasp-zap/Dockerfile
```diff
@@ -2,19 +2,33 @@ FROM python:3.8-alpine
...
-ARG AWS_CLI_VERSION="2.2.29"
-RUN curl -Lo awscliv2.zip "https://awscli.amazonaws.com/awscli-exe-linux-x86_64-${AWS_CLI_VERSION}.zip" \
-    && unzip awscliv2.zip \
-    && ./aws/install -i /usr/local/aws-cli -b /usr/local/bin
-
 COPY entrypoint.sh /entrypoint.sh
 
 # Launch OWASP scan
 ENTRYPOINT ["/entrypoint.sh"]
+# ENTRYPOINT ["tail", "-f", "/dev/null"]
```
dj2, line 34: Remove?

dj2 (approved):
lgtm w/ nit

runners/owasp-zap/Dockerfile
mohamed-cds: removed

PR #87, opened 2021-08-19
fix: address comments missed in an earlier PR, minor issues only

CalvinRodo:
# Summary | Résumé

Adds missing .PHONY targets
Add doc for lint-ci


api/Makefile
```diff
@@ -1,4 +1,4 @@
-.PHONY: dev fmt install lint migrations test fmt-ci
+.PHONY: dev fmt install lint migrations test fmt-ci lint-ci
```
dj2, line 1: nit: can you add `build install-dev` here as well?

dj2 (approved):
lgtm w/ nit.

patheard (approved):

PR #93, opened 2021-08-19
feat: Hot reload python files when running locally

### LINKED ISSUE #91: Hot reload python lambdas during dev.

CalvinRodo:
# Summary | Résumé

This PR does a couple of things.

First it reworks the Dockerfile for the API so that it runs expensive commands first so we can take advantage of caching.

For instance it will install playwright and chromium up front since those are slow and as it stands will always re-run whenever files change.

Secondly it installs the python modules in a different folder then the api code and sets the `PYTHONPATH` env variable to that folder so that we can mount the `api` folder as a volume mounted to `/function/` in the container.

Thirdly it installs `entr` and uses that to watch for changes to all *.py files in the api folder and subdirectories, when those files change it will restart aws-lambda-rie.

Closes #91

# Test instructions | Instructions pour tester la modification

Run the branch in a devcontainer, make a change to the python code, hit the endpoint and see the new code change occur.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -33,6 +33,6 @@
 	// Use 'forwardPorts' to make a list of ports inside the container available locally.
 	// "forwardPorts": [],
 	// Use 'postCreateCommand' to run commands after the container is created.
-	"postCreateCommand": "",
+	// "postCreateCommand": "",
 	"remoteUser": "vscode"
 }
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -33,6 +33,6 @@
 	// Use 'forwardPorts' to make a list of ports inside the container available locally.
 	// "forwardPorts": [],
 	// Use 'postCreateCommand' to run commands after the container is created.
-	"postCreateCommand": "pip3 install --user -r api/requirements.txt && pip3 install --user -r api/requirements_dev.txt && python -m playwright install chromium",
+	"postCreateCommand": "",
 	"remoteUser": "vscode"
 }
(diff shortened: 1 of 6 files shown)


.devcontainer/devcontainer.json
```diff
@@ -33,6 +33,6 @@
 	// Use 'forwardPorts' to make a list of ports inside the container available locally.
 	// "forwardPorts": [],
 	// Use 'postCreateCommand' to run commands after the container is created.
-	"postCreateCommand": "pip3 install --user -r api/requirements.txt && pip3 install --user -r api/requirements_dev.txt && python -m playwright install chromium",
+	"postCreateCommand": "",
```
patheard, line 36: nit - since this is now empty, you can smoke the line.

api/Dockerfile
```diff
@@ -10,23 +10,30 @@ RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
...
+
+RUN playwright install chromium
 
-RUN python3.9 -m pip install -r ${FUNCTION_DIR}/requirements.txt --target ${FUNCTION_DIR}
+COPY ./requirements.txt ${FUNCTION_DIR}
 
+
+RUN python3.9 -m pip install -r ${FUNCTION_DIR}/requirements.txt --target /pymodules
+
+RUN apt-get install entr
```
patheard, line 29: Do we need an `apt-get update` here?

patheard (commented):
Few minor comments, but it works oh so smooth for me locally.

api/Dockerfile
CalvinRodo: We update on line 20 I'll add entr to that line instead.

mohamed-cds (approved):
Worked like a charm 🚀 ✅

.devcontainer/devcontainer.json
CalvinRodo: done

api/Dockerfile
CalvinRodo: added it to other apt-get install

patheard (approved):

PR #99, opened 2021-08-24
feat: add template scan trigger model

maxneuvians:
This PR adds the `TemplateScanTrigger` model defined in #9. The only change was from the `callback:string` to `data:jsonb` field. Because we are not sure what that `callback` mechanism looks like, we can just keep it ambiguous as a `data` map.

### DIFF
--- a/api/db_migrations/versions/db98eafe1333_create_template_scan_triggers_table.py
+++ b/api/db_migrations/versions/db98eafe1333_create_template_scan_triggers_table.py
@@ -1,0 +1,35 @@
+"""create template scan triggers table
+
+Revision ID: db98eafe1333
+Revises: e6ec4f01db2f
+Create Date: 2021-08-24 17:21:59.537651
+
+"""
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+# revision identifiers, used by Alembic.
+revision = "db98eafe1333"
+down_revision = "e6ec4f01db2f"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.create_table(
+        "template_scan_triggers",
+        sa.Column("id", postgresql.UUID(as_uuid=True), primary_key=True),
+        sa.Column("template_scan_id", postgresql.UUID(as_uuid=True), nullable=False),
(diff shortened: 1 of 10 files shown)


api/db_migrations/versions/db98eafe1333_create_template_scan_triggers_table.py
```diff
@@ -0,0 +1,35 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.create_table(
+        "template_scan_triggers",
+        sa.Column("id", postgresql.UUID(as_uuid=True), primary_key=True),
+        sa.Column("template_scan_id", postgresql.UUID(as_uuid=True), nullable=False),
+        sa.Column("data", postgresql.JSONB(), nullable=False),
```
dj2, line 24: Should we continue to call it `callback` with type `jsonb` instead of `data`? My concern with data is it can be seen as a dumping ground of ... anything, which we probably don't want. Constraining the name makes it clear it's specifically related to the callback mechanism, whatever that is?

api/db_migrations/versions/db98eafe1333_create_template_scan_triggers_table.py
```diff
@@ -0,0 +1,35 @@
...
+        sa.Column("created_at", sa.DateTime, default=sa.func.now()),
+        sa.Column("updated_at", sa.DateTime, onupdate=sa.func.utc_timestamp()),
```
dj2, line 26: What's the difference between `now` and `utc_timestamp`?

maxneuvians: Probably one is TZ dependent, in the models themselves we use `default=datetime.datetime.utcnow,` so I updated `created_at` to `sa.func.utc_timestamp()`.

api/db_migrations/versions/db98eafe1333_create_template_scan_triggers_table.py
maxneuvians: Sounds good, change made!

dj2 (approved):

PR #100, opened 2021-08-24
feat: add scans model

maxneuvians:
This PR adds the `Scans` model as defined in #9. It also adds some more checks to existing model tests around missing relationship dependencies and makes sure order or import issues with the fixtures don't cascade into the tests.

### DIFF
--- a/api/db_migrations/versions/eb58f2b364a3_create_scans_table.py
+++ b/api/db_migrations/versions/eb58f2b364a3_create_scans_table.py
@@ -1,0 +1,40 @@
+"""create scans table
+
+Revision ID: eb58f2b364a3
+Revises: db98eafe1333
+Create Date: 2021-08-24 18:35:42.224567
+
+"""
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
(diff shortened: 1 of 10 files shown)


api/db_migrations/versions/eb58f2b364a3_create_scans_table.py
```diff
@@ -0,0 +1,40 @@
...
+        sa.Column("created_at", sa.DateTime, default=sa.func.utc_timestamp()),
+        sa.Column("updated_at", sa.DateTime, onupdate=sa.func.utc_timestamp()),
+        sa.ForeignKeyConstraint(
+            ["template_id"],
+            ["templates.id"],
+        ),
+        sa.ForeignKeyConstraint(
+            ["scan_type_id"],
+            ["scan_types.id"],
+        ),
```
dj2, line 35: Organization foreign key?

maxneuvians: Good catch - suddenly those model generators in Rails and Ecto make a lot more sense

dj2 (approved):

PR #101, opened 2021-08-24
feat: add A11yReport model

maxneuvians:
This PR adds the `A11yReport` model as defined in #9. It also adds better checks for JSONB fields in other models.

### DIFF
--- a/api/models/TemplateScan.py
+++ b/api/models/TemplateScan.py
@@ -42,5 +42,5 @@ class TemplateScan(Base):
 
     @validates("data")
     def validate_data(self, _key, value):
-        assert value != ""
+        assert value is not None
         return value
--- a/api/models/TemplateScan.py
+++ b/api/models/TemplateScan.py
@@ -43,4 +43,5 @@ class TemplateScan(Base):
     @validates("data")
     def validate_data(self, _key, value):
         assert value is not None
+        assert value != ""
         return value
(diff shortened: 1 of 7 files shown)


api/models/TemplateScan.py
```diff
@@ -42,5 +42,5 @@ class TemplateScan(Base):
 
     @validates("data")
     def validate_data(self, _key, value):
-        assert value != ""
+        assert value is not None
```
dj2, line 45: Does this still catch `""` ?

api/tests/models/test_A11yReport.py
```diff
@@ -0,0 +1,124 @@
...
+    assert a11y_report.revision == "revision"
+    assert a11y_report.url == "url"
+    assert_new_model_saved(a11y_report)
+    assert a11y_report.ci is False
+    assert a11y_report.summary == {"jsonb": "data"}
+    session.delete(a11y_report)
+    session.commit()
+
+
+def test_a11y_report_empty_productfails(scan_fixture, session):
```
dj2, line 61: Similar version of these tests by `product=""` instead of just missing?

maxneuvians: Yes those get caught:
```
    @validates("product")
    def validate_product(self, _key, value):
>       assert value != ""
E       AssertionError
```

api/models/TemplateScan.py
maxneuvians: I think that just becomes an empty json string, which is not deal either. I will add both tests

dj2 (approved):

PR #102, opened 2021-08-24
feat: bootstrapped basic web rendering using jinja2

mohamed-cds:
- Introduced Jinja and pydantic models so user input can be validated
- Ability to create an organization
- Dashboard view to list organizations
- Rudimentary error handling (needs to be refactored to common error handler)


dj2:
Before we land this, can we take it one step further to prove out i18n? We ran into this with the old site that integrating i18n caused us to throw out two templating systems where it was .... difficult to integrate. Would like to prove out the support before we decide on the library.

api/api_gateway/templates/dashboard.html
```diff
@@ -0,0 +1,27 @@
...
+  ​
+  <!-- Compiled and minified CSS -->
+  <link href="https://unpkg.com/tailwindcss@^2/dist/tailwind.min.css" rel="stylesheet" />
+  <title>Scan Websites</title>
+</head>
+​
+<body>
+  {% block content %}
+  Organizations
+  <tr>
```
patheard, line 18: This would be better as a `<ul>`:
```
 <ul> 
  {% for organization in organisations %}
    <li>{{organization.name}}</li>
  {% endfor %}
  </ul>
```

api/api_gateway/templates/page.html
```diff
@@ -0,0 +1,220 @@
...
+  <meta charset="UTF-8" />
+  <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+  <meta http-equiv="X-UA-Compatible" content="ie=edge" />
+  ​
+  <!-- Compiled and minified CSS -->
+  <link href="https://unpkg.com/tailwindcss@^2/dist/tailwind.min.css" rel="stylesheet" />
+  <title>Scan Websites</title>
+  <style>
+    /*remove custom style*/
+    .circles {
```
patheard, line 15: It would be good to get this in a custom `.css` file for easier sharing/maintenance across pages.

api/api_gateway/templates/dashboard.html
```diff
@@ -0,0 +1,27 @@
...
+
+<head>
+  <!--Import Google Icon Font-->
+  <meta charset="UTF-8" />
+  <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+  <meta http-equiv="X-UA-Compatible" content="ie=edge" />
+  ​
+  <!-- Compiled and minified CSS -->
+  <link href="https://unpkg.com/tailwindcss@^2/dist/tailwind.min.css" rel="stylesheet" />
+  <title>Scan Websites</title>
```
patheard, line 12: Have we thought about how we're going to do translation?

api/api_gateway/api.py
```diff
@@ -30,3 +36,32 @@ async def healthcheck():
...
+
+@app.get("/", response_class=HTMLResponse)
+async def home(request: Request):
+    data = {"page": "Home page"}
+    return templates.TemplateResponse("page.html", {"request": request, "data": data})
+
+
+@app.post("/organization", response_class=HTMLResponse)
+async def create_organization(organization: OrganizationBase):
+    session = db_session()
```
dj2, line 49: Can we add a todo for authentication here?

api/api_gateway/api.py
```diff
@@ -30,3 +36,32 @@ async def healthcheck():
...
+    try:
+        new_organization = Organisation(name=organization.name)
+        session.add(new_organization)
+        session.commit()
+    except Exception as e:
+        raise HTTPException(status_code=500, detail=str(e))
+
```
dj2, line 56: Where does this go on success? Should it redirect to the organization page for the new org?

api/api_gateway/api.py
```diff
@@ -30,3 +36,32 @@ async def healthcheck():
...
+
+@app.get("/dashboard", response_class=HTMLResponse)
+async def dashboard(request: Request):
+    session = db_session()
```
dj2, line 60: todo auth?

api/api_gateway/api.py
```diff
@@ -30,3 +36,32 @@ async def healthcheck():
         db_status = {"able_to_connect": False}
 
     return {"database": db_status}
+
+
+@app.get("/", response_class=HTMLResponse)
+async def home(request: Request):
+    data = {"page": "Home page"}
+    return templates.TemplateResponse("page.html", {"request": request, "data": data})
```
dj2, line 44: If I'm signed in, should this auto-redirect to dashboard? If so, add a todo?

api/api_gateway/api.py
```diff
@@ -30,3 +36,32 @@ async def healthcheck():
...
+        session.commit()
+    except Exception as e:
+        raise HTTPException(status_code=500, detail=str(e))
+
+
+@app.get("/dashboard", response_class=HTMLResponse)
+async def dashboard(request: Request):
+    session = db_session()
+    try:
+        result = session.query(Organisation).all()
```
dj2, line 62: This probably needs to limit to the organization for the current user.

api/api_gateway/api.py
```diff
@@ -30,3 +36,32 @@ async def healthcheck():
...
+
+@app.post("/organization", response_class=HTMLResponse)
+async def create_organization(organization: OrganizationBase):
+    session = db_session()
+    try:
+        new_organization = Organisation(name=organization.name)
+        session.add(new_organization)
+        session.commit()
+    except Exception as e:
+        raise HTTPException(status_code=500, detail=str(e))
```
dj2, line 55: Will this write out the exception in prod as well? We probably want to log the exception and write a friendly error string. (And other exceptions below)

api/api_gateway/templates/dashboard.html
```diff
@@ -0,0 +1,27 @@
+<!DOCTYPE html>
+<html>
```
dj2, line 2: This needs a `lang` tag

api/api_gateway/templates/dashboard.html
```diff
@@ -0,0 +1,27 @@
...
+
+<head>
+  <!--Import Google Icon Font-->
```
dj2, line 5: This comment is unrelated to the code below

api/api_gateway/templates/page.html
```diff
@@ -0,0 +1,220 @@
+<!DOCTYPE html>
```
dj2, line 1: Does jinja have layout templates? We shouldn't need to repeat the header stuff for each page should we?

api/api_gateway/templates/page.html
```diff
@@ -0,0 +1,220 @@
...
+        border-radius: 50%;
+      }
+
+    }
+  </style>
+</head>
+​
+
+<body>
+  <section class="text-gray-700 body-font"></section>
```
dj2, line 131: This all seems like it could be removed?

api/schemas/Organization.py
```diff
@@ -0,0 +1,9 @@
+from pydantic import BaseModel
+
+
+class OrganizationBase(BaseModel):
```
dj2, line 4: What does this provide over `api/modules/Organisation.py`?

mohamed-cds: Pydantic is used for documentation, validation, and data serialization. SQLAlchemy for SQL ORM stuff

api/api_gateway/templates/page.html
mohamed-cds: This PR won't be merged until I have a working template that supports french translations and is accessible. This is placeholder for now to test the ORM

api/schemas/Organization.py
dj2: Can we add this into the other model file? Having 2 different models for the same thing seems ... confusing? When do I use OrganizationBase and when do I use Organization? When am I dealing with a model vs schema?

mohamed-cds: It's main purpose is to allow us to filter results back from api. e.i remove password from user object, as well as validate user input so that extraneous data isn't sent to the API that it doesn't expect. Makes it easier to code review since ORM models should never be sent back to the user when using pydantic.

I seperated it out since thats how the pydantic docs have it. Don't see a problem with moving the code to the models.

api/api_gateway/api.py
mohamed-cds: Added todo. Need to figure out how auth will work before that goes in

api/database/db.py
```diff
@@ -3,7 +3,10 @@
 from sqlalchemy import create_engine
 from sqlalchemy.orm import sessionmaker
 
+# from models import initialize_sql
```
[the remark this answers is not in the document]
mohamed-cds: removed

api/front_end/templates/base.html
```diff
@@ -0,0 +1,57 @@
...
+  <body class="bg-gray-100 font-sans leading-normal tracking-normal h-full">
+    <div class="min-h-screen flex flex-col">
+      <div class="flex-grow">
+        <nav class="bg-white w-full z-10 top-0 shadow" id="header">
+          <div class="w-full container mx-auto flex flex-wrap items-center mt-0 pt-3 pb-3 md:pb-0">
+            <div class="pl-2 pb-4 w-1/2 md:pl-0">
+              <a class="text-gray-900 text-base no-underline font-bold xl:text-xl hover:no-underline" href="/{{ lang }}" aria-label="{{ goc }}">
+                <img class="h-6 inline-block pr-10" src="{{ goc_banner }}" alt="{{ goc }}" />
+              </a>
+              <span class="leading-none lg:ml-4 mr-5 lg:pl-5 font-semibold inline-block top-0 text-brand pb-0">Scan Websites</span>
```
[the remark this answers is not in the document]
mohamed-cds: done :)

api/front_end/templates/base.html
```diff
@@ -0,0 +1,57 @@
...
+              <span class="bg-blue-200 py-1 px-2 rounded-lg text-small">ALPHA</span>
```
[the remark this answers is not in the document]
mohamed-cds: alpha is the same in both english & french, but i added locale anyways for consistency and ease of change in the future.

api/front_end/templates/base.html
```diff
@@ -0,0 +1,57 @@
...
+                {{ other_language }}
+              </a>
+            </div>
+          </div>
+        </nav>
+        <div class="container w-full mx-auto h-full">
+          <div class="w-full px-4 mb-16 text-gray-800 leading-normal md:px-0 md:mt-8">
+            <div class="flex flex-col flex-1 max-h-full pl-2 pr-2 rounded-md xl:pr-4">
+              <main class="flex-1 pt-2">
+                {% block body %}{% endblock %}
```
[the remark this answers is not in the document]
mohamed-cds: removed content since it wasn't necessary anymore

api/api_gateway/v1/api.py
```diff
@@ -0,0 +1,76 @@
...
+from logger import log
+
+from models.Organisation import Organisation
+from schemas.Organization import OrganizationCreate
+
+# from crawler.crawler import crawl
+# import uuid
+from pydantic import BaseModel
+
+app = FastAPI()
```
maxneuvians, line 16: we may need that root path attribute to make the docs site work

maxneuvians (approved):
Nice work - let us try it out and see if we have any gaps! Small comment on the docs site.

api/api_gateway/v1/api.py
mohamed-cds: Going to refactor versioning to use https://github.com/DeanWay/fastapi-versioning in a separate PR and will add the root path

PR #106, opened 2021-09-02
chore: refactored to organize routes and client

mohamed-cds:
Refactored the api so that components reside in their own route. This is needed to allow unique routes so that we don't clobber existing routes when using similar paths.

Also added a dev flag for the root_path so that we have consistency between local test and what's deployed to AWS. this can be removed once we migrate to scan-websites.alpha.canada.ca


maxneuvians (approved):
Nice! Thank you for including bandit!

PR #112, opened 2021-09-07
feat: gsuite login and route protection

### LINKED ISSUE #110: Add Single Sign On (SSO)
Currently there is no way to restrict access to content by organization. The initial implementation of SSO should be to only allow Gsuite authentication and eventually expanded to a solution that can be used by other government departments.

mohamed-cds:
Closes #110. 

Changes
- Login via Gsuite implemented. Can only be used by cds-snc google workspace members.
- Database will be seeded with a CDS org and a accessibility + security scan types
- password_hash in user database now allows null since users can login via gsuite

### DIFF
--- a/api/db_migrations/versions/5818f4679595_seed_databases.py
+++ b/api/db_migrations/versions/5818f4679595_seed_databases.py
@@ -43,13 +43,13 @@ def upgrade():
         [
             {
                 "id": str(uuid.uuid4()),
-                "name": "Accessibility - Accessibilité",
+                "name": "axe-core",
                 "created_at": datetime.now(),
                 "updated_at": datetime.now(),
             },
             {
                 "id": str(uuid.uuid4()),
-                "name": "Security - Sécurité",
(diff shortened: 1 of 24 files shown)


api/db_migrations/versions/5818f4679595_seed_databases.py
```diff
@@ -0,0 +1,70 @@
...
+                "updated_at": datetime.now(),
+            }
+        ],
+    )
+
+    op.bulk_insert(
+        scan_types,
+        [
+            {
+                "id": str(uuid.uuid4()),
```
maxneuvians, line 45: Lets figure out if this should be an over arching category "Security" or "ZAP", "DNS" etc

mohamed-cds: Changed to "OWASP Zap" & "axe-core"

maxneuvians (approved):
Really nice work!

PR #113, opened 2021-09-08
feat: store axe reports

### LINKED ISSUE #63: Store A11yReport
Based on the model proposed in #9, save the axe-core report to the DB. We can assume that each report will have the following top level keys:

```
{
"url": "ex. https://digital.canada.ca",
"id": "UUIDv4",
"report": { ... }
}
```

with the URL and ID being injected through the crawler here: https://github.com/cds-snc/scan-websites/blob/main/api/crawler/crawler.py#L34 and here: https://github.com/cds-snc/scan-websites/blob/main/api/pub_sub/pub_sub.py#L16

Currently there is no way to directly inject a URL and ID into the axe-core pub-sub from the API that is not the crawler and something that needs to be addressed.

This issue should also extract the content of the report and save the relevant data as a summary. An example of that summary can be found here https://github.com/cds-snc/a11y-tools/blob/main/processing/persist-lambda/src/impl.ts#L148-L156 although this should inspire more than anything else.

Attached is a sample JSON of the an axe-core report.

[sample.json.txt](https://github.com/cds-snc/scan-websites/files/6983243/sample.json.txt)

maxneuvians:
Closes #63. This PR adds the ability to break apart axe core reports from an S3 event and then store them in the DB as `A11yReport` and `A11yViolation` objects.

It also changes the PubSub component so that when a URL is sent to be scanned, it creates a A11yReport object to indicate that a scan is in progress. Once that is completed and sent back through the bucket, it will update the status. This way you can see if A11yReports are being started but not completed.

It also adds a quality of life configuration that will speed up tests by setting the bcrypt rounds to 4 from 14 in CI.


api/models/User.py
```diff
@@ -43,7 +46,9 @@ def password(self):
 
     @password.setter
     def password(self, password):
```
mohamed-cds, line 48: if we're going down the sso path; do we even need a password_hash anymore ? Or do you think we'll have username/password for non-cds folks until we find a more global sso option implemented?

maxneuvians: I would be happy to eliminate it - it will force us to figure out how to implement SSO for non CDS folks :)

mohamed-cds: i'll be setting password_hash to "nullable" in my upcoming PR. This way we can decide what to do with form login at a later date!

api/db_migrations/versions/5818f4679595_seed_databases.py
```diff
@@ -62,7 +62,10 @@ def downgrade():
     op.alter_column("users", "password_hash", nullable=False)
 
     op.execute(
-        """DELETE FROM "organisations" WHERE name = 'Canadian Digital Service - Service Numérique Canadien'; """
+        """
+        DELETE FROM users;
```
dj2, line 66: Won't this cause issues for anyone that's already migrated the seed_databases command?

api/pub_sub/pub_sub.py
```diff
@@ -1,20 +1,35 @@
...
+        url=payload["url"],
+        summary={"state": "scanning"},
+        scan=scan,
+    )
+    session.add(a11y_report)
+    session.commit()
+    session.close()
 
+    # Add a ID that can be linked back to the parent ID of the payload
+    payload["id"] = a11y_report.id
```
dj2, line 32: Does alchemy do the auto-reload of the id into the object or does it have to be queried again?

api/pub_sub/pub_sub.py
```diff
@@ -1,20 +1,35 @@
...
     # Currently only dispatches to axe-core
     AXE_CORE_URLS_TOPIC = os.environ.get("AXE_CORE_URLS_TOPIC", False)
 
-    # Add a ID that can be linked back to the parent ID of the payload
-    payload["id"] = str(uuid.uuid4())
+    a11y_report = A11yReport(
+        product="product",  # TODO how is this populated?
+        revision="revision",  # TODO how is this populated?
+        url=payload["url"],
+        summary={"state": "scanning"},
```
dj2, line 24: Later we use `status`, here we use `state` can this be `status`?

api/tests/storage/test_storage.py
```diff
@@ -39,5 +49,94 @@ def test_get_object_catch_exception(mock_get_session, mock_log):
...
+        "violations": {"moderate": 2, "serious": 1, "total": 3},
+    }
+
+
+def test_store_axe_core_record_creates_violations(session, scan_fixture):
+    a11y_report = A11yReport(
+        product="product",
+        revision="revision",
+        url="url",
+        summary={"jsonb": "data"},
```
dj2, line 116: This summary doesn't seem to match the other formats?

api/storage/storage.py
```diff
@@ -16,3 +23,73 @@ def get_object(record):
             f"Error downloading {record.s3.object.key} from {record.s3.bucket.name}"
         )
         return False
+
+
+def store(record):
+    name = record.s3.bucket.name
+    body = get_object(record)
```
dj2, line 30: This feels weird that store has to get an object first. It would be clearer if the object was passed into `store` instead of being retrieved.

api/storage/storage.py
```diff
@@ -16,3 +23,73 @@ def get_object(record):
...
+
+    if not body:
+        return False
+
+    try:
+        payload = json.loads(body)
+
+        if name == os.environ.get("AXE_CORE_REPORT_DATA_BUCKET", None):
```
dj2, line 38: Can this check be negated and moved up to line 29 so it happens immediately and we exit as early as possible?

api/storage/storage.py
```diff
@@ -16,3 +23,73 @@ def get_object(record):
...
+
+    except Exception:
+        log.error(f"Error decoding {record.s3.object.key} from {name}")
+        return False
+
+
+def store_axe_core_record(payload):
+    session = db_session()
+    a11y_report = session.query(A11yReport).get(payload["id"])
+    if a11y_report is not None:
```
dj2, line 52: Can this be negated and do an early exit?

api/storage/storage.py
```diff
@@ -16,3 +23,73 @@ def get_object(record):
...
+                            target=node["target"],
+                            html=node["html"],
+                            data=item["data"],
+                            tags=violation["tags"],
+                            message=item["message"],
+                            url=payload["url"],
+                            a11y_report=a11y_report,
+                        )
+                        session.add(a11y_violation)
+                        session.commit()
```
dj2, line 81: Is the commit here fast or would be be beneficial to move the commit out of the `for` loops and just do one commit for all the violations?

api/db_migrations/versions/5818f4679595_seed_databases.py
maxneuvians: No it is just the downgrade. Because there is not SQL defined CASCADE the downgrade fails if there are users in the table.

api/pub_sub/pub_sub.py
maxneuvians: Yea it gets autorefreshed

api/pub_sub/pub_sub.py
maxneuvians: Good catch!

api/tests/storage/test_storage.py
maxneuvians: It can just be any JSONB, we have not existing constraints

api/storage/storage.py
maxneuvians: Yes

api/storage/storage.py
maxneuvians: Probably the later - moved it

api/storage/storage.py
dj2: That's starting to feel like it's overloading the storage class. Why does storage care if it's an axe_core report, or some_other_report? It should just be talking a place and storing it there. The processing should be done up  front, no?

Maybe my abstraction is set at a different point, what does `storage` mean?

maxneuvians: I think it just decides how to store it? Storing happens in the type specific functions anyways - so instead `retrieve_and_store` it could be `retrieve_and_route`? I hate naming things :(

dj2: Ah, that makes sense, because the store requires breaking it apart to fit into the db tables. For some reason my brain got stuck thinking store like store an s3 blob, so why would we care what's inside it. But this is store individual db records.

dj2 (approved):

mohamed-cds (approved):
LGTM 🚀 ✅

PR #126, opened 2021-09-13
fix: route root path requests to the lambda api

### LINKED ISSUE #124: Bug: Browsing the root directory of the website displays an error
The API gateway needs a new root resource "/" added and a "ANY" method so that requests to the root directory of the https://scan-websites.alpha.canada.ca/ website is routed to the fastapi lambda.

mohamed-cds:
Closes #124. The api gateway will now route / requests to the lambda for processing. Should result in the index page being loaded.


maxneuvians (approved):
🦑

PR #130, opened 2021-09-14
feat: refactor owasp-zap to be consistent with axe-core

mohamed-cds:
Setup for processing owasp zap reports similar to how axe-core reports are ingested.


maxneuvians (approved):
We can remove PRIVATE_API_AUTH_TOKEN if you think it makes sense

PR #133, opened 2021-09-16
Minor formatting and cleanup of various files.

dj2:
# Summary | Résumé

This does some cleanup to the various Docker and other files in the
project.


dj2:
PTAL. The Zap dockerfile is the biggest change I think? As far as I understand it should be equivalent, but .....

mohamed-cds (approved):
Don't see any changes to existing behaviour, thanks for the cleanup! LGTM 🚀 ✅

PR #135, opened 2021-09-19
feat: add Terragrunt to `Lint` GitHub workflow

### LINKED ISSUE #132: Terraform: add checkov security scan GitHub Workflow
The majority of the fixes should be the same as those applied to the ListManager:
* cds-snc/list-manager#80
* cds-snc/list-manager#81

patheard:
# Summary 
* Add the `terragrunt` folder to the matrix jobs to lint and format check the code.
* Add a check for changes within each matrix folder to exit early if there's no new code.

**Note:** although it would work to remove the `matrix` and use the root level `Makefile` to execute the workflow steps, using a matrix approach is faster as each folder will run in parallel.

Related #132

### DIFF
--- a/.github/workflows/ci_code.yml
+++ b/.github/workflows/ci_code.yml
@@ -7,6 +7,7 @@ on:
       - scanners/axe-core/**
       - scanners/owasp-zap/**
       - terragrunt/**
+      - .github/workflows/ci_code.yml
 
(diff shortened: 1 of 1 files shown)


mohamed-cds (approved):
nice! 🚀 ✅

PR #138, opened 2021-09-27
fix: move hsts functionality from api gateway to application layer

mohamed-cds:
First attempt to [set hsts at the api gateway](https://github.com/cds-snc/scan-websites/runs/3725043252) failed due to proxy integrations not allowing transforming responses. `Proxy integrations cannot be configured to transform responses`

This PR will move that functionality to fastapi via custom middleware.

### DIFF
--- a/api/api_gateway/custom_middleware.py
+++ b/api/api_gateway/custom_middleware.py
@@ -1,0 +1,12 @@
+from starlette.middleware.base import BaseHTTPMiddleware
+
+
+class HSTSHeaderMiddleware(BaseHTTPMiddleware):
+    def __init__(self, app, header_value="max-age=300; includeSubDomains; preload"):
+        super().__init__(app)
+        self.header_value = header_value
+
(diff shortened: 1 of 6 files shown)


api/api_gateway/custom_middleware.py
```diff
@@ -0,0 +1,7 @@
+from fastapi import Request
+
+
+async def add_security_headers(request: Request, call_next):
+    response = await call_next(request)
+    response.headers["Strict-Transport-Security"] = "max-age=300 ; includeSubDomains"
```
patheard, line 6: Just because I'm curious, why no `preload` option this time?

patheard (approved):
Small comment, but not merge blocking.

api/api_gateway/custom_middleware.py
mohamed-cds: omitted until we've confirmed the behaviour. I'll add it back once we set max-age to 1 year. Didn't want google caching our settings

PR #141, opened 2021-10-01
feat: dynamically start a11y and security scans using tokens

mohamed-cds:
Callback added to scan_types so each scan type can have its associated callback/trigger associated with it. Template scan triggers will now be used to trigger scans configured to run at a frequency instead of being triggered by API (all templates can be manually triggered).


api/api_gateway/routers/scans.py
```diff
@@ -56,6 +60,85 @@ def template_belongs_to_org(
...
+        raise HTTPException(status_code=status.HTTP_401_UNAUTHORIZED)
+
+    if user.organisation_id != template.organisation_id:
+        raise HTTPException(status_code=status.HTTP_401_UNAUTHORIZED)
+
+    return template
+
+
+@router.get("/start")
+@limiter.limit("1/minute")
```
maxneuvians, line 91: Would this be a global limit? A per user limit? or is this just a temporary safe guard?

api/pub_sub/pub_sub.py
```diff
@@ -6,31 +6,48 @@
...
+            url=payload["url"],
+            summary={"status": "scanning"},
+            scan=scan,
+        )
+        session.add(a11y_report)
+        session.commit()
+        payload["id"] = str(
+            a11y_report.id
+        )  # Add a ID that can be linked back to the parent ID of the payload
+        session.close()
```
patheard, line 45: Could this be generalized or do we need one `if` block per scan type (e.g. will they be different enough going forward that generalizing is a waste of time)?

And, grain of salt, I usually go too far down the DRY road...

api/api_gateway/routers/scans.py
mohamed-cds: will be a global limit for each unique IP. 1/minute was copy/paste from /crawl. Thinking of 10/minute for this one. This can be removed once we have something upstream doing the same thing.

api/pub_sub/pub_sub.py
mohamed-cds: the results go to different tables, so we have to keep them separate. As we create more scan types, we can potentially generalize the results tables.

patheard: Makes sense - I had been thinking more around the code generalization so we didn't end up with a long `if/if else/if else` block, but it might not be awful in practice.  Definitely in the "future us" category of problems.

api/api_gateway/routers/scans.py
mohamed-cds: removed since Pat added the firewall!

api/pub_sub/pub_sub.py
mohamed-cds: 100% agree. I just need to see where this goes before refactoring

api/models/TemplateScan.py
```diff
@@ -38,7 +39,13 @@ class TemplateScan(Base):
...
     scan_type = relationship("ScanType", back_populates="template_scans")
 
-    template_scan_triggers = relationship("TemplateScanTrigger")
+    template_scan_trigger_id = Column(
+        UUID(as_uuid=True),
+        ForeignKey(TemplateScanTrigger.id),
+        index=True,
+        nullable=True,
+    )
+    # template_scan_trigger = relationship("TemplateScanTrigger", foreign_keys=[template_scan_trigger_id])
```
patheard, line 48: nit: delete commented code

api/tests/pub_sub/test_pub_sub.py
```diff
@@ -7,17 +7,150 @@
 @patch("pub_sub.pub_sub.A11yReport")
 @patch("pub_sub.pub_sub.db_session")
 @patch("pub_sub.pub_sub.send")
-@patch.dict(os.environ, {"AXE_CORE_URLS_TOPIC": "topic"}, clear=True)
+# @patch.dict(os.environ, {"AXE_CORE_URLS_TOPIC": "topic"}, clear=True)
```
patheard, line 10: Can we remove this commented code as well?

api/models/TemplateScan.py
mohamed-cds: thanks for catching this! Meant to bring it back and was only temporarily commented out

patheard (approved):

maxneuvians (approved):

Issue #111, opened 2021-09-07
Replace slowapi rate limiting implementation with WAF ruleset once available


mohamed-cds:
Rate limit introduced [here](https://github.com/cds-snc/scan-websites/pull/109). 

Implementation example AWS WAF rate limit.

https://stackoverflow.com/questions/65252674/when-using-terraform-with-aws-how-can-i-set-a-rate-limit-on-a-specific-uri-path

patheard:
Related #142

PR #144, opened 2021-10-04
feat: add WAF ACL for API

### LINKED ISSUE #142: Terraform: add WAF rules for API
The API should be protected by a WAF ACL.  

Related #132

patheard:
# Summary
Associate a Web Application Firewall Access Control List with
the API to protect it from malicious input.

Also provides a rate limit of 100 requests from the same IP in a
5 minute period (the minimum allowed by a rate limit statement).
If these needs to be lower, we can add multiple rate limit statements.

Related #111
Closes #142


mohamed-cds (approved):
🚀 ✅  fingers crossed we don't trigger any of the rulesets!

PR #145, opened 2021-10-04
fix: exclude AWS common rule

patheard:
# Summary
The `GenericRFI_QUERYARGUMENTS` rule is blocking login with Google accounts.

Related #142

### DIFF
--- a/terragrunt/aws/api/waf.tf
+++ b/terragrunt/aws/api/waf.tf
@@ -45,6 +45,10 @@ resource "aws_wafv2_web_acl" "api_waf" {
         excluded_rule {
           name = "GenericRFI_BODY"
         }
+
+        excluded_rule {
+          name = "GenericRFI_QUERYARGUMENTS"
+        }
       }
     }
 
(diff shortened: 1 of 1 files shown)


CalvinRodo:
What specifically is being sent as a query param that is being blocked? 

Also I'm curious if we have something in place that will protect us from the attacks we are potentially opening ourselves to by disabling this rule, or if we even care for this product. 

FYI The rule we are disabling does this:

> Inspects the value of query arguments and blocks common cross-site scripting (XSS) patterns using the built-in XSS detection rule in AWS WAF. Example patterns include scripts like <script>alert("hello")</script>.

patheard:
It's failing on the `://` in the query params for the Google account sign in callback:
```sh
/auth/google?state=${STATE}&code=${CODE}&scope=email%20profile%20openid%20https://www.googleapis.com/auth/userinfo.email%20https://www.googleapis.com/auth/userinfo.profile&authuser=0&hd=cds-snc.ca&prompt=none
```

> Inspects the values of all query parameters and blocks requests attempting to exploit RFI (Remote File Inclusion) in web applications. Examples include patterns like ://.

Currently there's nothing else protecting against this in the app.

Related, we may not need `GenericRFI_BODY` excluded.  That was a copy/paste pull-in from ListManager that can probably be removed.

CalvinRodo (approved):
Lets give it a shot.

:taco::cat:

PR #147, opened 2021-10-05
feat: make template components visible in the client

mohamed-cds:
Can now view your organisation, it's templates, and then configure scans for each template.


CalvinRodo (approved):
:taco::cat:

api/front_end/templates/user.html
```diff
@@ -0,0 +1,26 @@
+{% extends "base.html" %}
+{% block includes %}
+<script src="{{ url_for('static', path='js/user.js') }}"></script>
+{% endblock %}
+{% block body %}
+  
+    <div class="container flex justify-center items-center">
+      <div>
+        <h1>{{ request.session.user.name }}</h1>
+        <p><b>{{ email_locale }}:</b> {{ request.session.user.email_address }}</p>
```
patheard, line 10: Minor tweak: should use `<strong>` instead of `<b>`

api/front_end/templates/user.html
```diff
@@ -0,0 +1,26 @@
...
+        <div id='child'>
+        <b>{{ api_key_locale }}:</b>
```
patheard, line 12: Same comment re: `<strong>`

patheard (changes requested):
Two minor changes in the HTML markup, but otherwise, good stuff!

patheard:
One other quest I have that's probably outside the scope of this PR is can we standardize on FontAwesome or Google Material icons?

patheard (approved):

mohamed-cds:
> One other question I have that's probably outside the scope of this PR is can we standardize on FontAwesome or Google Material icons?

Removed FontAwesome since i was only using it for the google icon which was barely visible!

PR #150, opened 2021-10-05
feat: allow api to decrypt kms used for sns topics

mohamed-cds:
Grant the API access to kms in-order to publish to encrypted sns topic. Applied locally to test


CalvinRodo (approved):
:taco::cat:

PR #151, opened 2021-10-06
feat: template view now allows multiple scans per template

mohdnr:
Previous limit of one scan per template has been removed since new database model allows it. The ability to delete scans has also been implemented.

### DIFF
--- a/api/front_end/static/js/template.js
+++ b/api/front_end/static/js/template.js
@@ -1,8 +1,8 @@
 $(document).ready(function() {
   $(document).on("click", "#deleteScan a", function(event) {
     event.preventDefault();
-    event.stopPropagation();
-      if (confirm($(this).attr('data-confirm'))) {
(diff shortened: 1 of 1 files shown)


api/front_end/static/js/template.js
```diff
@@ -0,0 +1,22 @@
+$(document).ready(function() {
+  $("#deleteScan a").click(function(event) {
```
patheard, line 2: A tweak you can make here is to change how the click event is bound:
```js
$(document).on("click", "#deleteScan a", function(event) {
```
That way you won't run into problems with trying to bind events for elements that don't exist in the DOM at bind time.

patheard (commented):
Had a few comments around the JS, but all minor.

api/front_end/static/js/template.js
mohamed-cds: Changed :)

PR #159, opened 2021-10-12
feat: github compliance scanner

CalvinRodo:
# Summary | Résumé

This PR adds a scanner that will scan github for compliance with organizational defined best-practices.

This scanner is triggered by an SNS topic.

This scanner has permission to write to an s3 bucket a json report of the configuration of the app, so we can determine if it's compliant with CDS Policies.


bin/scanner_github_test.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+
+echo 
+#curl "http://scanner-github:8080/2015-03-31/functions/function/invocations" -d '{ }'
```
patheard, line 4: nit: remove commented code

terragrunt/aws/scanners/github/lambda.tf
```diff
@@ -0,0 +1,63 @@
...
+    effect = "Allow"
+
+    actions = [
+      "logs:CreateLogGroup",
+      "logs:CreateLogStream",
+      "logs:PutLogEvents"
+    ]
+
+    resources = [
+      "*"
```
patheard, line 30: Could this be scoped down to just the Lambda log group?

terragrunt/aws/scanners/github/lambda.tf
```diff
@@ -0,0 +1,63 @@
...
+
+    actions = [
+      "s3:ListBucket",
+      "s3:ListBucketVersions",
+      "s3:GetBucketLocation",
+      "s3:Get*",
+      "s3:Put*"
+    ]
+    resources = [
+      "*"
```
patheard, line 59: Same comment about scoping down if possible.

patheard (commented):
Had a few minor comments, but looks good.  I like the simplicity of how the compliance check query is managed.

terragrunt/aws/scanners/github/lambda.tf
CalvinRodo: This should be in the lambda module.

bin/scanner_github_test.sh
CalvinRodo: Done

patheard (approved):

PR #165, opened 2021-10-13
fix: set playwright path

maxneuvians:
This PR sets a playwright path in the API docker build. When the image is built chromium is installed in a different path than it can be accessed by the user executing the lambda. This should create a hermetic build.


mohamed-cds (approved):
🧑‍🔬 ✅

PR #184, opened 2021-10-18
feat: run sqli tests seperately and use only 1 thread

mohdnr:
[Multi-threaded scans cause OWASP Zap to flag SQLi injection issue falsely](https://github.com/zaproxy/zaproxy/issues/3662). The workaround is to exclude sqli tests from the multithreaded scan and to test for sqli separately using 1 thread. The impact primarily being a small increase in scan times.

This PR also introduces the capability to toggle the number of threads via terraform input variable.

### DIFF
--- a/runners/owasp-zap/entrypoint.sh
+++ b/runners/owasp-zap/entrypoint.sh
@@ -42,9 +42,10 @@ zap-cli --port "${ZAP_PORT}" --zap-url "http://$HOST_IP" ajax-spider "${SCAN_URL
 # Gets list of available scans, excludes sqli and converts to csv
 all_except_sqli=$(zap-cli --zap-url http://"$HOST_IP" --port "${ZAP_PORT}" scanners list | tail -n +3 | grep -v "SQL Injection" | awk -F ' ' '{print $2}' | awk NF | awk '$1=$1' RS= OFS=,)
 
+# Set scan threads as defined by environment variable
 curl "http://${HOST_IP}:${ZAP_PORT}/JSON/ascan/action/setOptionThreadPerHost/" -H 'Content-Type: application/x-www-form-urlencoded' --data-raw "Integer=${SCAN_THREADS}" --compressed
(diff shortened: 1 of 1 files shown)


runners/owasp-zap/entrypoint.sh
```diff
@@ -38,8 +38,22 @@ zap-cli --port "${ZAP_PORT}" --zap-url "http://$HOST_IP" open-url "${SCAN_URL}"
...
 zap-cli --port "${ZAP_PORT}" --zap-url "http://$HOST_IP" ajax-spider "${SCAN_URL}"
 
+# Initial scan will exclude sqli since false positivies occur when sqli is run multithreaded
+# Gets list of available scans, excludes sqli and converts to csv
+all_except_sqli=$(zap-cli --zap-url http://"$HOST_IP" --port "${ZAP_PORT}" scanners list | tail -n +3 | grep -v "SQL Injection" | awk -F ' ' '{print $2}' | awk NF | awk '$1=$1' RS= OFS=,)
+
+# Set scan threads as defined by environment variable
+curl "http://${HOST_IP}:${ZAP_PORT}/JSON/ascan/action/setOptionThreadPerHost/" -H 'Content-Type: application/x-www-form-urlencoded' --data-raw "Integer=${SCAN_THREADS}" --compressed
+
+# Timeout scan after 2 hour to prevent running indefinitely if the OWASP ZAP container crashes
```
patheard, line 48: Out of curiosity, why the 2 hour timeout for this scan, but the 1 hour timeout for the `sqli` scan below?

mohamed-cds: the first active-scan is for multi scan types; cross-site scripting, template injection ,etc. While the 2nd one which is set to 1 hour only scans for SQL Injection and is expected to complete sooner.

It used to be 1 hour but i increased the scope of what's being scanned beyond the defaults.

PR #196, opened 2021-10-18
feat: delete report button

mohamed-cds:
Ability to delete reports

### DIFF
--- a/api/front_end/templates/template_scan_list.html
+++ b/api/front_end/templates/template_scan_list.html
@@ -33,7 +33,7 @@
                 </td>
                 <td class="px-6 py-4 whitespace-nowrap text-right text-sm font-medium">
                   <a href="/{{lang}}/template/{{ template_scan.template.id }}/scan/{{ template_scan.id}}" class="text-green-600 hover:text-indigo-900"><span class="material-icons" aria-label="{{ edit_locale }}"> edit </span></a>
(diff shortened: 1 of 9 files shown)


api/front_end/templates/template_scan_list.html
```diff
@@ -33,7 +33,7 @@
                 </td>
                 <td class="px-6 py-4 whitespace-nowrap text-right text-sm font-medium">
                   <a href="/{{lang}}/template/{{ template_scan.template.id }}/scan/{{ template_scan.id}}" class="text-green-600 hover:text-indigo-900"><span class="material-icons" aria-label="{{ edit_locale }}"> edit </span></a>
-                  <span id="deleteScan"><a href="/scans/template/{{ template_scan.template.id }}/scan/{{ template_scan.id}}" data-confirm="{{scan_delete_confirm_locale}}" class="text-red-600 hover:text-indigo-900"><span class="material-icons" aria-label="{{ delete_locale }}"> delete </span></a></span>
+                  <span id="deleteConfirm"><a href="/scans/template/{{ template_scan.template.id }}/scan/{{ template_scan.id}}" data-confirm="{{scan_delete_confirm_locale}}" class="text-red-600 hover:text-indigo-900"><span class="material-icons" aria-label="{{ delete_locale }}"> delete </span></a></span>
```
patheard, line 36: Out of curiosity, why not just put the `#deleteConfirm` ID on the `<a>` element?
```html
<a id="deleteConfirm">
```
And then adjust the event binding:
```js
$(document).on("click", "#deleteConfirm", // etc.
```
There's nothing wrong with what you have, but my preference has always been to limit the number of HTML elements.

mohamed-cds: Sounds good to me! Made the change

PR #200, opened 2021-10-19
fix: add missing model so site renders

mohdnr:
All models need to be listed in main.py otherwise sqlalchemy cries....


patheard (approved):
The _last_ thing I would want to do is upset SQLAlchemy.

PR #205, opened 2021-10-20
feat: seed local database to make testing easier

mohdnr:
Ability to seed local database with templates, reports and results


maxneuvians (approved):
❤️ that you did it with the models vs. raw SQL

PR #221, opened 2021-11-02
refactor: move to one lambda and use stepfunctions to launch scans

### LINKED ISSUE #170: Bump HSTS from 1 week (604800) to 1 month (2592000)
The [recommendation](https://hstspreload.org/#deployment-recommendations) when implementing HSTS is to gradually introduce the change to confirm behaviour before setting it to 1,2 years + preload.

After this issue is closed, another issue should be created to bump this to one year and preload.

mohamed-cds:
Instead of one lambda per long running scan via ECS. This PR will migrate towards a scan "orchestrator" that will launch scans via step functions.

The benefit of step functions is that it allows scans to be run sequentially while getting deduplication natively. This setup will also remove the need to create a dedicate lambda launcher for each ecs task.

Closes #170


patheard (approved):
Super interested to see this in action.

PR #223, opened 2021-11-02
fix: task def unique family

mohamed-cds:
Tested locally

### DIFF
--- a/runners/nuclei/entrypoint.sh
+++ b/runners/nuclei/entrypoint.sh
@@ -1,5 +1,11 @@
 #!/bin/sh -l
 
+trap 'aws stepfunctions send-task-failure --task-token "${TASK_TOKEN_ENV_VARIABLE}" --task-output "Caught SIGTERM signal"' TERM
+trap 'aws stepfunctions send-task-failure --task-token "${TASK_TOKEN_ENV_VARIABLE}" --task-output "Caught SIGINT signal"' INT
+
+heartbeat_cmd="while sleep 60; do aws stepfunctions send-task-heartbeat --task-token ${TASK_TOKEN_ENV_VARIABLE}; done"
+eval "$heartbeat_cmd" &
+
 nuclei --version
 
 date=$(date +\"%Y-%m-%dT%H:%M:%S:%N\")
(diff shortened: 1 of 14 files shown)


runners/nuclei/entrypoint.sh
```diff
@@ -1,5 +1,11 @@
 #!/bin/sh -l
 
+trap 'aws stepfunctions send-task-failure --task-token "${TASK_TOKEN_ENV_VARIABLE}" --task-output "Caught SIGTERM signal"' TERM
+trap 'aws stepfunctions send-task-failure --task-token "${TASK_TOKEN_ENV_VARIABLE}" --task-output "Caught SIGINT signal"' INT
+
+heartbeat_cmd="while sleep 60; do aws stepfunctions send-task-heartbeat --task-token ${TASK_TOKEN_ENV_VARIABLE}; done"
```
patheard, line 6: Is there any chance of this infinitely looping?  I'm just wondering if there should be a break condition after some period of time that represents "something has gone weird".

runners/owasp-zap/entrypoint.sh
```diff
@@ -3,6 +3,9 @@
 trap 'aws stepfunctions send-task-failure --task-token "${TASK_TOKEN_ENV_VARIABLE}" --task-output "Caught SIGTERM signal"' TERM
 trap 'aws stepfunctions send-task-failure --task-token "${TASK_TOKEN_ENV_VARIABLE}" --task-output "Caught SIGINT signal"' INT
 
+heartbeat_cmd="while sleep 60; do aws stepfunctions send-task-heartbeat --task-token ${TASK_TOKEN_ENV_VARIABLE}; done"
```
patheard, line 6: Same question about infinite loops.

runners/nuclei/entrypoint.sh
mohamed-cds: Handled by the `TimeoutSeconds: 7200,` that's been added to the task with the understanding that the background task will not prevent the container from exiting. End condition will be added if that turns out to be false.

runners/owasp-zap/entrypoint.sh
mohamed-cds: Handled by the `TimeoutSeconds: 7200,` that's been added to the task with the understanding that the background task will not prevent the container from exiting. End condition will be added if that turns out to be false.

patheard (approved):

PR #230, opened 2021-11-05
feat: api launch scanning state machine

mohamed-cds:
🧪 Launch step functions from the API so that the lambda can be removed.


maxneuvians (approved):
For 🧪

PR #238, opened 2021-11-05
feat: remove unused scan starter lambda

mohamed-cds:
Now that dynamic scans have been moved to step functions, the scan starter lambda is no longer required.


patheard (approved):
Tidy 🧹

PR #241, opened 2021-11-08
feat: toggle dyanmic scans; disabled by default

mohamed-cds:
Toggle dynamic security scans using the dynamic query parameter. Disabled by default since dynamic scans will not be run as frequently as static scans.


maxneuvians (approved):
💣 🎉

PR #251, opened 2021-12-07
feat: add aws otel lambda as docker image 

dinophile:
Also included: 

update gitignore  

add otel tracing variable to lambda tf

# Summary | Résumé

Added code from AWS Otel lambda layer to docker image in ecr. We then pull that image in to run it directly in the app since layers are not available for lambdas using containers. (Yet, presumably they will add this function à la Honeycomb et al.)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -200,3 +200,6 @@ dist
 
 # Ignore local stack data file
 .devcontainer/data
+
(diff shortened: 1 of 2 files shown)


mohamed-cds (approved):
🧪 Let's give this a try!  ✅

PR #252, opened 2021-12-08
fix: add AWS Lambda exec wrapper env variable

patheard:
# Summary
Re-enables the AWS Lambda exec wrapper environment variable to use the `otel-instrument` binary to execute requests.


dinophile (approved):
:shipit:

PR #262, opened 2021-12-15
feat: delete a11y report

mohamed-cds:


api/front_end/templates/scan_results_a11y.html
```diff
@@ -82,8 +88,13 @@ <h1 class="text-gray-700 font-semibold uppercase text-xl mb-7">{{ scan.template.
                         <a href="{{a11y_report.url}}" target="_blank">
```
patheard, line 88: An a11y improvement you could make here is to add an `aria-label` to describe what the link is doing:
```html
<a href="{{a11y_report.url}}" aria-label="View a11y report" target="_blank">
```

patheard (changes requested):
Looks good!  

Had a minor comment about an a11y improvement and you'll also have to remove out the lambda API's env variable that's trying to run the OTel stuff that has been removed from the docker image:
https://github.com/cds-snc/scan-websites/blob/main/terragrunt/aws/api/lambda.tf#L14

patheard (approved):

PR #264, opened 2021-12-20
feat: retrieve lambda layer on build

mohamed-cds:
Instead of using a prebuilt docker image, download the contents of the lambda layer instead.


dinophile (approved):
🍽️

PR #265, opened 2021-12-20
feat: use custom otel config with debug enabled

mohamed-cds:
Custom config so that debug can be enabled.


dinophile (approved):
Ahhh there they are! 🧙

PR #267, opened 2021-12-21
fix: working opentelemetry

mohamed-cds:
AWS lambda opentelemetry auto instrumentation assumes that you're setup exactly the same as lambda runtime environments (either using lambda layers or container images).

Migrating to using the amazon/lambda-python will make it less work to maintain observability in the future or to leverage new tools that work in lambda.


dinophile (approved):
:shipit:

PR #269, opened 2021-12-21
fix: dont exit the loop when skipping, actually skip

mohamed-cds:
Candidate for the most embarrassing bug of 2021 for @mohamed-cds


CalvinRodo (approved):
:taco::Cat:

PR #270, opened 2021-12-22
feat: debian based opentelemetry autoinstrumentation

mohamed-cds:
The missing link to the autoinstrumentation of our original docker image was the call to the python wrapper defined using the `AWS_LAMBDA_EXEC_WRAPPER` environment variable. This bootstraps the lambda handler invocation so that opentelemetry can create begin and end spans when the lambda starts and to detect all loaded python libraries that support auto instrumentation.

image


dinophile (approved):
🌭

PR #271, opened 2021-12-22
fix: background no longer blocked by middleware

mohamed-cds:
When the security headers middleware was introduced that caused the background tasks to no longer run in the background since it was being blocked by the middleware. This is due to https://github.com/encode/starlette/issues/919. Moving towards a custom middleware class that uses pure ASGIApp resolves the issue.

Using this class also fixed the issue were the middleware wasn't working well in the CI environment and needed customization in-order to run tests.

Background tasks should work as expected and an immediate response received when scans are started.


maxneuvians (approved):
🎉 Nice work!

PR #277, opened 2022-01-10
feat: add roles for tf plan and apply using OICD

CalvinRodo:
# Summary | Résumé

- Adds a role restriced to main branch that can has the admin managed
  policy attached.

- Adds a role that has no branch restriction that has the readonly
  managed policy attached

**Please Note:** I'm not 100% sure the readonly role will work so testing in a low risk repo. Also this is part 1 of 2 planned PR's the second will be changing the TF actions.


mohamed-cds (approved):
for science 🧪 ✅

patheard (approved):

PR #282, opened 2022-01-11
feat: use OpenID Connect to auth TF actions

CalvinRodo:
# Summary | Résumé

Authenticates the Plan action using a readonly account and apply using
an admin account.

### DIFF
--- a/.github/workflows/tf_plan.yml
+++ b/.github/workflows/tf_plan.yml
@@ -6,9 +6,6 @@ on:
       - "terragrunt/**"
       - ".github/workflows/**"
 env:
-  AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
-  AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
-  AWS_REGION: ca-central-1
   TERRAFORM_VERSION: 1.0.3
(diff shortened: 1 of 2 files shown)


.github/workflows/tf_plan.yml
```diff
@@ -59,6 +59,13 @@ jobs:
           chmod +x bin/*
           echo "$GITHUB_WORKSPACE/bin" >> $GITHUB_PATH
 
+      - name: configure aws credentials using OIDC
+        uses: aws-actions/configure-aws-credentials@master
+        with:
+          role-to-assume: arn:aws:iam::507252742351:role/gh_plan_role
+          role-session-name: TFApply
```
patheard, line 66: Since this is the plan job, would it make sense to have this as `TFPlan`?  I'm assuming this shows up somewhere in an audit log?

patheard (commented):
Good stuff - had a question about the `role-session-name` on the plan workflow but I'm looking forward to copy/pasting this all over the place.

.github/workflows/tf_plan.yml
CalvinRodo: It would make sense, it shows up in a list of active sessions and would show up in the audit log

patheard (approved):

patheard (approved):
Lets giver.  I'm curious if the TF apply will give weird errors about missing permissions on the workflow, but only one way to find out.

PR #289, opened 2022-01-28
feat: add Cloud Based Sensor integration

mohamed-cds:
This will send VPC flow logs and WAF logs to the CBS satellite bucket in the account.

### DIFF
--- a/terragrunt/aws/api/waf.tf
+++ b/terragrunt/aws/api/waf.tf
@@ -1,6 +1,4 @@
 resource "aws_wafv2_web_acl" "api_waf" {
-  # TODO: Create kinesis stream and logging config
-  # checkov:skip=CKV2_AWS_31:Kinesis required,will do in seperate PR
   name        = "api_waf"
   description = "WAF for API protection"
   scope       = "REGIONAL"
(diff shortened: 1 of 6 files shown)


terragrunt/aws/api/waf.tf
```diff
@@ -205,3 +205,41 @@ resource "aws_wafv2_web_acl_association" "waf_association" {
...
+
+resource "aws_cloudwatch_log_group" "api_waf" {
+  name              = "/aws/kinesisfirehose/api_waf"
+  retention_in_days = 14
+
+  tags = {
+    CostCentre = var.billing_code
+    Terraform  = true
+  }
+}
```
patheard, line 219: I was just debating exactly this on the ListManager PR I'm working on.  Do you want to add a CloudWatch log group source as well to the Kinesis stream?  

It would make it easier to debug WAF requests, but there'd be the extra cost of the data retention (no idea how much though).

terragrunt/aws/api/waf.tf
```diff
@@ -205,3 +205,41 @@ resource "aws_wafv2_web_acl_association" "waf_association" {
...
+
+resource "aws_kinesis_firehose_delivery_stream" "api_waf" {
```
patheard, line 221: I think we should add encryption at rest with a block like so:
```hcl
server_side_encryption {
  enabled = true
}
```

patheard (approved):
science! 🧪

PR #291, opened 2022-01-31
fix: firehose write to s3 missing iam permissions

mohamed-cds:
Adds missing IAM policies that was preventing WAF ACL logs from being written to S3

### DIFF
--- a/terragrunt/aws/api/waf.tf
+++ b/terragrunt/aws/api/waf.tf
@@ -211,7 +211,7 @@ resource "aws_kinesis_firehose_delivery_stream" "api_waf" {
 
   extended_s3_configuration {
     role_arn   = aws_iam_role.waf_log_role.arn
-    prefix     = "waf_acl_logs"
+    prefix     = "waf_acl_logs/"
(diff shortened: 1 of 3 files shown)


terragrunt/aws/api/waf.tf
```diff
@@ -211,7 +211,7 @@ resource "aws_kinesis_firehose_delivery_stream" "api_waf" {
 
   extended_s3_configuration {
     role_arn   = aws_iam_role.waf_log_role.arn
-    prefix     = "waf_logs"
+    prefix     = "waf_acl_logs"
```
patheard, line 214: Could you add a trailing `/` to this.  I was reading a docs and that's what causes it to write items to their own "folder" in the bucket.

patheard (changes requested):
Just a minor change for the prefix.

Also, I was reading about the CloudWatch config we had been talking about on Friday and it's actually CloudWatch config for error logging the S3 object writes, so it's the type of thing we could add if we ever needed to debug why logs weren't showing up in the bucket.

patheard (approved):

PR #292, opened 2022-01-31
chore: add prefix for vpc flow logs

mohamed-cds:
Test vpc flow logs prefix. [According to AWS logs it should be possible to a prefix when defining the bucket name](https://docs.aws.amazon.com/vpc/latest/userguide/flow-logs-s3.html)


patheard (approved):
🧪

PR #296, opened 2022-02-03
fix: UTC time was causing inconsistent filenames

mohamed-cds:
S3 was automatically encoding the `:` in the UTC time of filenames, resulting in errors when the api attempts to retrieve axe-core report data. Moving from UTC time to epoch time as a unique identifier will ensure no special characters exist in filenames.

Also upgrading aws-sdk from 2.1054.0 to 2.1055.0 to resolve a Snyk reported vulnerability.

### DIFF
--- a/scanners/axe-core/package-lock.json
+++ b/scanners/axe-core/package-lock.json
@@ -10,7 +10,7 @@
       "license": "MIT",
       "dependencies": {
         "@axe-core/puppeteer": "^4.3.1",
-        "aws-sdk": "^2.1046.0",
+        "aws-sdk": "^2.1055.0",
         "puppeteer": "^10.2.0"
       },
       "devDependencies": {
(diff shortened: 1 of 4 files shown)


patheard (commented):
Code change looks good to me, but I guess this highlights that our `terraform plan/apply` workflows needs to be adjusted since otherwise we won't get a deployment of this fix.

Or is there a different deploy mechanism for this project?

mohamed-cds:
> Code change looks good to me, but I guess this highlights that our `terraform plan/apply` workflows needs to be adjusted since otherwise we won't get a deployment of this fix.
> 
> Or is there a different deploy mechanism for this project?

@patheard  `build_and_push.yml` will take care of deploying to ecr. TF not required since the lambda will get updated.

patheard (approved):

PR #298, opened 2022-02-09
fix: validate before referencing optional fields

mohamed-cds:
Nuclei output has some optional outputs that are not consistent. This PR will validate before trying to store some of the optional fields.


patheard (commented):
Is there a way to add a unit test for this validation in `api/storage/storage.py`?

patheard (approved):

PR #305, opened 2022-03-05
feat: adds account source to lambda invoke permissions

maxneuvians:
This PR adds the source account to the lambda invoke permissions as an additional safeguard to invoking.


mohamed-cds (approved):
Thanks for taking care of this!! 🚀 ✅

PR #311, opened 2022-04-08
feat: common workflow to launch scans

mohamed-cds:
Re-useable "start scan" workflow so that onboarded products all stay in sync if the process to start scans change in the future.


CalvinRodo (approved):
:taco::cat:

PR #313, opened 2022-05-06
refactor: template creation

mohamed-cds:
Refactored the template creation process to no longer accept key/value pairs from users. I've also introduced input fields that will accept urls that should not be scanned.


dinophile (approved):
Nice clean up!

PR #319, opened 2022-05-19
fix: sync with clickops

maxneuvians:
We clickopsed a severless v2 database and now would like to see how the TF reconciles that


CalvinRodo (approved):
:taco::cat:

PR #320, opened 2022-05-20
feat: add docker sbom generate 

patheard:
# Summary
1. Generate SBOMs for the Docker images.
2. Update to use the SBOM GitHub action rather than the shared workflow.
3. Adjust the triggers on the Terraform workflows so they only run if there are infrastructure code changes.

# Related
* cds-snc/platform-sre-security-support#94


mohamed-cds (approved):
this will be a nice example when onboarding other teams. 🚀 ✅

PR #344, opened 2022-06-17
* Add platform enforcement to support ARM chips.

### LINKED ISSUE #343: .devcontainer - Platform Dependent Terragrunt Hash
With Devs having differing hardware, need to tweak the Terragrunt download/hash to be variable based on the platform of the Dev. 

Shouldn't default to just amd64.

See TERRAGRUNT_CHECKSUM in docker-compose.yml

ShadeWyrm:
# Summary | Résumé

Closes #343

Sets the platform for the Docker Containers to amd64 so that platforms running arm64 don't try to override the platform (and thus don't work)

# Test instructions | Instructions pour tester la modification

Run Dev env on arm64 device, and it will no longer be mad about dependencies that aren't available on arm64.

### DIFF
--- a/.devcontainer/docker-compose.yml
+++ b/.devcontainer/docker-compose.yml
@@ -6,6 +6,7 @@ x-db-connection-strings: &db-connection-strings
 
 services:
   app:
+    platform: linux/x86_64
     build:
       context: ..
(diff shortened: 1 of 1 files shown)


mohamed-cds (approved):
LGTM 🚀 ✅

PR #350, opened 2022-07-07
fix: reworked devcontainer setup to not have conflicting path configs

mohdnr:
The dockerfile was setting PATH variables in two locations of which the first had precedence and was overriding all other attempts to modify the PATH environment variable.

I've centralized the configuration of the PATH environment variable to the `devcontainer.json` file

Bonus: Re-introduced the postcreate command so that the devcontainer will automatically perform the `make install & make install-dev` commands when a new devcontainer is built. I suspect this was commented out due to the issues related to the PATH env.


patheard (approved):
Good stuff!

PR #370, opened 2022-09-08
chore: upgrade RDS to postgres 14.3

patheard:
# Summary
Upgrade RDS cluster to latest `aurora-postgresql` engine version.

Also adds a quality of life fix to the devcontainer so that Terraform code
is formatted on save.


dinophile (approved):
✅

PR #371, opened 2022-09-20
fix: update database ACU and major version

patheard:
# Summary
Update the Serverless V2 ACUs so that upgrades can be performed properly.

Also updates to Postgres `14.3`.


CalvinRodo (dismissed):
:taco::cat:

maxneuvians (approved):

PR #480, opened 2023-04-03
fix: reverting changes

ascheid:
# Summary | Résumé

Changes made to this code were implemented in the shared config renovate app. PR to stop overriding the changes that are now part of that config.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,10 +1,5 @@
 {
-  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
   "extends": [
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -1,10 +1,5 @@
 {
-  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
```
patheard, line 2: You need this `$schema` key

patheard: Seems this is not the case.

